PR #24, opened 2020-01-21
Implement auth password reset

IdezHD:
Password reset completed minus the sending of emails, temp page in place

### DIFF
--- a/web/track/views.py
+++ b/web/track/views.py
@@ -13,7 +13,7 @@ from flask import render_template, Response, abort, request, redirect, url_for
 from flask_login import LoginManager, login_required, login_user
 from flask_bcrypt import Bcrypt
 from track import models
-from flask_login import LoginManager, login_required
+from flask_login import LoginManager, login_required, current_user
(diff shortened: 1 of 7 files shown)


web/track/views.py
```diff
@@ -372,51 +373,62 @@ def forgot_password():
...
             if user_password == user_password_confirm:
                 if is_strong_password(user_password):
                     # Create a user to insert into the database
-                    # to_add = Users(
-                    #     user_password=bcrypt.generate_password_hash(user_password),  # Flask-Bcrypt password hash
-                    # )
-                    # connection.insert(to_add)
-                    # connection.commit()
-                    return render_template(generate_path(prefix, "password-changed"))
+                    user_password = bcrypt.generate_password_hash(user_password),  # Flask-Bcrypt password hash
```
peacheym: Since our project is based on Python 3, we need to add a .decode(‘utf-8’) to the end of this hashing function call.

https://flask-bcrypt.readthedocs.io/en/latest/#usage Take a look here if you'd like!

PR #30, opened 2020-01-22
Password reset now includes email functionality

IdezHD:
Notification api has been successfully tested with a new email template

### DIFF
--- a/web/track/views.py
+++ b/web/track/views.py
@@ -396,7 +396,10 @@ def register(app):
 
             user_email = cleanse_input(request.form.get('email_input'))
             user = connection.query_user_by_email(user_email)
-            send_pass_reset(user_email, user, prefix)
+
+            if user is not None and user.is_authenticated:
+                template_id = '8c3d96cc-3cbe-4043-b157-4f4a2bbb57b1'
+                send_pass_reset(user, prefix, template_id)
 
             msg = 'If an account is associated with the email address, ' \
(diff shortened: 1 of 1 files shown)


web/track/views.py
```diff
@@ -487,20 +486,22 @@ def verify_account_mobile():
         phone = 'placeholder: ***-***-3333'
         return render_template(generate_path(prefix, "verify-mobile"), phone=phone)
 
-
     def send_pass_reset(user_email, user, prefix):
+        user_name = user.display_name
```
peacheym: Instead of assigning the user.display_name to a variable, can we just put it directly in the personalisation dict? Just make line 500 -> 'user': user.display_name, ?? :)

web/track/views.py
```diff
@@ -487,20 +486,22 @@ def verify_account_mobile():
...
                                          _external=True
                                          )
-            return render_template(generate_path(prefix, "reset-email-temp"), password_reset_url=password_reset_url)
-            # response = notifications_client.send_email_notification(
-            #     email_address='email_address',
-            #     template_id='6e3368a7-0d75-47b1-b4b2-878234e554c9'we
-            # )
-            # General Notification
+            response = notifications_client.send_email_notification(
+                email_address=user_email,
```
peacheym: Can we just get the user_email from the user obj being passed into this function instead of passing redundant params?

PR #39, opened 2020-01-24
Remove __init__.py from types

sleepycat:
This file make pipenv choke on trying to import name 'MethodType' from
'types', since the types folder was considered a module.


IdezHD:
This should no longer be required with updates to API

PR #40, opened 2020-01-27
Flask graphql models

IdezHD:
- Created individual schemas for each model located in the schemas directory
- Created models file which contains all models found in postgres db
- Created db file to handle creation of database connection


api/README.md
```diff
@@ -0,0 +1,17 @@
+<h1>API</h1>
```
peacheym: It would probably be best to use Markdown for the readme file.  It will be a quick port over!

api/models.py
```diff
@@ -1,18 +1,167 @@
...
+class Users(base):
+    __tablename__ = 'users'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    username = Column(String)
+    display_name = Column(String)
+    user_email = Column(String)
+    user_password = Column(String)
+    preferred_lang = Column(String)
+    failed_login_attempts = Column(Integer)
```
peacheym: May be worth re-adding the `default=0` for failed_login_attempts

PR #43, opened 2020-01-28
GraphQL create new users through mutation

peacheym:
Allows for new users to be created and stored in postgres via the graphql api.
Adds a mutation for generating an JWT Auth token.
TODO:  Move he Bcrypt instance into a global scope.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -5,7 +5,7 @@ from sqlalchemy import Column, String
 from sqlalchemy.orm import relationship
 from sqlalchemy.dialects.postgresql import JSONB
 
-from db import Base as base
+from .db import Base as base
 
 
 class Domains(base):
--- a/api/models.py
+++ b/api/models.py
@@ -38,6 +38,7 @@ class Organizations(base):
     domains = relationship("Domains", back_populates="organization", cascade="all, delete")
     users = relationship("User_affiliations", back_populates="user_organization", cascade="all, delete")
 
+
 class Groups(base):
     __tablename__ = 'groups'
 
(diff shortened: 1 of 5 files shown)


api/models.py
```diff
@@ -5,7 +5,7 @@
 from sqlalchemy.orm import relationship
 from sqlalchemy.dialects.postgresql import JSONB
 
-from db import Base as base
+from .db import Base as base
```
IdezHD: The full stop is not required anymore

api/queries.py
```diff
@@ -1,13 +1,16 @@
-import graphene
-from graphene import relay
-from graphene_sqlalchemy import SQLAlchemyConnectionField
-
-from schemas.user import UserConnection
+from .schemas.auth_token import *
```
IdezHD: Full stops no longer required

api/schemas/user.py
```diff
@@ -1,17 +1,60 @@
...
+from flask_graphql_auth import create_refresh_token
+from graphene import relay, Mutation
+from graphene_sqlalchemy import SQLAlchemyObjectType, SQLAlchemyConnectionField
 
-from models import Users as UserModel
+from flask_bcrypt import Bcrypt
+
+from flask import current_app as app
+
+from ..models import Users as UserModel
```
IdezHD: Full stops not required

api/schemas/user.py
```diff
@@ -1,17 +1,60 @@
...
+	create_access_token,
+	create_refresh_token,
+	query_jwt_required,
+	mutation_jwt_refresh_token_required,
+	mutation_jwt_required,
+)
 
 
 class User(SQLAlchemyObjectType):
 	class Meta:
```
IdezHD: Class should have to following exclude_fields = 'user_password', this ensures that no query from outside can access the passwords columns

PR #45, opened 2020-01-28
Add graphql schema export script

sleepycat:
This script introspects on the GraphQL schema and writes the results
into schema.json. This description can later be used as the basis for
comparing schemas between frontend and api service to ensure that they
stay in sync.

When Graphql core 3 is more widely supported this script can be [made much cooler](https://graphql-core-next.readthedocs.io/en/latest/usage/introspection.html).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 **/.DS_Store
 
+api/schema.json
+
 **/node_modules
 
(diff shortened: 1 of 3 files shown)


IdezHD:
Looks pretty interesting, and should be super useful when developing!

PR #47, opened 2020-01-29
Graph ql sign up in functions

IdezHD:
- Added error checking into the create user mutation
- Created signing in functionality, with error checking
- Added query property to all models to allow querying of each model

### DIFF
--- a/api/queries.py
+++ b/api/queries.py
@@ -8,8 +8,7 @@ class Query(graphene.ObjectType):
 
 
 class Mutation(graphene.ObjectType):
-	create_user = CreateUser.Field()
-	# auth = AuthMutation.Field()
+	create_user = CreateUser.Field()=
 	sign_in = SignInUser.Field()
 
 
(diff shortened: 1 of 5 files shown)


api/queries.py
```diff
@@ -9,7 +9,8 @@ class Query(graphene.ObjectType):
 
 class Mutation(graphene.ObjectType):
 	create_user = CreateUser.Field()
-	auth = AuthMutation.Field()
+	# auth = AuthMutation.Field()
```
peacheym: It looks like you have moved the access token into the new sign-in endpoint.   Can this commented out line be removed?

api/schemas/user.py
```diff
@@ -57,4 +69,40 @@ def mutate(self, info, username, password, email):
...
+	@classmethod
+	def mutate(cls, _, info, email, password):
+		email = cleanse_input(email)
+		password = cleanse_input(password)
+		user = User.query.filter(User.user_email == email).first()
+
+		if user is None:
+			raise GraphQLError("User does not exist, please register")
+
+		bcrypt = Bcrypt(app)
```
peacheym: Does having separate bcrypt objects cause issues for generating/verifying passwords?  If not this PR is good to go!

IdezHD: No it doesn't affect any operations

PR #50, opened 2020-01-29
Update Error Messages & Add unit tests

peacheym:
This PR adds unit tests for password based api calls.

This PR also refactors error messages into their own functions file so that a single function can be used for both api code as well as testing code.

### DIFF
--- a/api/schemas/user.py
+++ b/api/schemas/user.py
@@ -1,15 +1,10 @@
 import graphene
-import json
-from graphene import relay, Mutation
+from graphene import relay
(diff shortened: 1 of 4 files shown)


api/schemas/user.py
```diff
@@ -4,6 +4,9 @@
 from graphene_sqlalchemy import SQLAlchemyObjectType
 
 from models import Users as User
+from db import db_session
```
IdezHD: Lines 7,8, and 9 are no longer needed. Only needed in the function files

PR #54, opened 2020-01-30
Graph ql creating initial scalars

IdezHD:
Also added in a new regex


api/tests/test_email_address_scalar.py
```diff
@@ -0,0 +1,67 @@
...
+PACKAGE_PARENT = '..'
+SCRIPT_DIR = os.path.dirname(os.path.realpath(os.path.join(os.getcwd(), os.path.expanduser(__file__))))
+sys.path.append(os.path.normpath(os.path.join(SCRIPT_DIR, PACKAGE_PARENT)))
+
+from scalars.email_address import *
+
+
+class TestEmailAddressScalar(unittest.TestCase):
+
+	def testValidEmailSerialize(self):
```
peacheym: I believe that functions should follow snake_case conventions rather than camelCase.

api/tests/test_email_address_scalar.py
```diff
@@ -0,0 +1,67 @@
...
+	def testValidEmailParseValue(self):
+		test_email = "test.account@canada.ca"
+		assert EmailAddress.parse_value(test_email)
+
+	def testValidEmailParseLiteral(self):
+		assert EmailAddress.parse_literal(ast.StringValue(
+			value="test.account@canada.ca"
+		))
+
+	def testInvalidEmailSerialize1(self):
```
peacheym: I also think it would be in our best interest to name functions something more descriptive than <something>1 and <something>2!

PR #92, opened 2020-02-13
Setup Azure db connection for creating enums when testing, updated api/cloudbuild to accommodate changes

IdezHD:
- Created system to accommodate the use of Azure DB to create the enums for testing purposes


api/model_enums/__init__.py
```diff
@@ -22,4 +30,5 @@ def create_enum_app():
 def create_enum_db(enum_app):
 	enum_db = SQLAlchemy()
 	enum_db.init_app(enum_app)
+
 	return enum_db
```
peacheym: Looks like this file is missing a new line at the end -- also looks like it wasn't you that caused it!

api/model_enums/__init__.py
```diff
@@ -2,14 +2,22 @@
...
+
 
 def create_enum_app():
 	enum_app = Flask(__name__)
 
-	DB_USER = os.getenv('DB_USER')
-	DB_PASS = os.getenv('DB_PASS')
-	DB_HOST = os.getenv('DB_HOST')
-	DB_NAME = os.getenv('DB_NAME')
+	if _called_from_test:
```
peacheym: As long as Mike is okay with us calling it like this!

PR #94, opened 2020-02-13
Implement basic user roles functionality

peacheym:
# Adds basic user role functionality
This PR's largest role is to enable pulling from the `user_role` column from the **Users** table in the database and adding it to the `user_claims` dictionary when signing in a user.

### Additions
Adds a `update_role()` function that allows an **admin** to update the role of a user.
Adds a `test_user_claims()` function that allows you to hit the API and see what a user's role is.

Both of the above additions are protected by the `flask_graphql_auth` decorators.


api/manage.py
```diff
@@ -8,11 +8,12 @@
...
 	DB_HOST,
 	DB_PASS,
-	DB_USER
+	DB_USER,
+	DB_PORT,
 )
 
 app = Flask(__name__)
-app.config['SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}/{DB_NAME}'
+app.config['SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
```
IdezHD: Just wondering why you added the port, the system will automatically go for the correct port?

peacheym: I added it because I used a non-default port for my local postgres instance.  This is to avoid all conflicts between my local and my cloud-build-local.  Plus, were passing it in as an ENV so we may as well use it?  Unless of course it will break something to do with migrate?

PR #98, opened 2020-02-14
Graphql organizations resolvers

IdezHD:
- Created auto generation of enums with descriptions
- Implemented Organization enums, via auto generation
- Implemented Organizations initial resolvers 
- Fixed resolver tests to work with auto generation of enums through Azure

### DIFF
--- a/api/functions/generate_enums.py
+++ b/api/functions/generate_enums.py
@@ -1,4 +1,5 @@
 import graphene
+import enum
 from sqlalchemy.orm import load_only
 from model_enums import create_enum_app, create_enum_db
 from models import (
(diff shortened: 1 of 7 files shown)


api/functions/generate_enums.py
```diff
@@ -40,9 +41,54 @@ def create_enums(Table, enum_name, column):
...
-		return graphene.Enum(enum_name, 'EMPTY')
+		ret_enum = enum.Enum(enum_name, 'EMPTY')
 	else:
-		return graphene.Enum(enum_name, enum_dict)
+		ret_enum = enum.Enum(enum_name, enum_dict)
+
+	return graphene.Enum.from_enum(ret_enum)
+
+
+def create_enum_with_descriptions(Table, enum_name, enum_column, description_column):
```
peacheym: I've been adding the whole `:param:` and `:returns:` to the doc strings -- may be worth it to do the same!  :)

api/functions/generate_enums.py
```diff
@@ -40,9 +41,54 @@ def create_enums(Table, enum_name, column):
...
+
+	for e in ret_enum:
+		with app.app_context():
+			query = db.session.query(Table).filter(getattr(Table, enum_column) == e.name).options(load_only(description_column))
+			row = db.session.execute(query).first()
+
+			try:
+				description_dict.update({e.name: row[1]})
+			except:
+				description_dict.update({e.name: "Error loading description"})
```
peacheym: Could this go with the other error messages in `functions/error_messages.py`??

api/queries.py
```diff
@@ -70,6 +78,24 @@ class Query(graphene.ObjectType):
...
+		of_type=Organizations,
+		id=graphene.Argument(graphene.Int, required=True),
+		resolver=resolve_get_org_by_id,
+		description="Allows the selection of an organization from a given ID"
+	)
+	get_org_by_org = graphene.List(
+		of_type=Organizations,
+		org=graphene.Argument(OrganizationsEnum, required=True),
+		resolver=resolve_get_org_by_org,
+		description="Allows the selection of an organization from its given organization code"
```
peacheym: the 'given' feels redundant in the description.  It's early though :yum:

api/resolvers/organizations.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Resolvers
+def resolve_get_org_by_id(self, info, **kwargs):
+	"""Return an organization by its id"""
+	org_id = kwargs.get('id')
+	query = Organizations.get_query(info).filter(
+		OrganizationsModel.id == org_id
+	)
+	if not len(query.all()):
+		raise GraphQLError("Error, Invalid ID")
```
peacheym: Can all of the error messages go into functions in this file?  Easier to manage them all in one spot!

api/tests/test_organizations_resolver.py
```diff
@@ -0,0 +1,213 @@
...
+						description
+						groupId
+					}
+				}
+			"""
+			executed = client.execute(query)
+
+		assert executed['errors']
+		assert executed['errors'][0]
+		assert executed['errors'][0]['message'] == "Error, Invalid ID"
```
peacheym: This will be able to be a function call also if we move them into `functions/error_messages.py`!!

api/functions/generate_enums.py
IdezHD: This should not go into a function, if there is an error and the entire description column is empty and you have 1000+ rows that you are processing it will call that function the 1000+ on startup, this way it just sets a direct value

api/resolvers/organizations.py
IdezHD: It's a specific and short error message doesn't need to be put into a function, and it will never change

api/tests/test_organizations_resolver.py
IdezHD: ^^^

PR #101, opened 2020-02-14
Add more functionalty and tests to user-roles

peacheym:
## User roles becoming more robust


api/functions/update_user_role.py
```diff
@@ -2,6 +2,7 @@
 from db import db
 from models import Users as User
 from graphql import GraphQLError
+from user_roles import *
```
IdezHD: Is it possible to be more specific with imports, sometimes when importing all you can overwrite changes that are required that occurred in other files

api/resolvers/users.py
```diff
@@ -1,7 +1,7 @@
 from graphql import GraphQLError
 from flask_graphql_auth import *
-
-from functions.error_messages import error_not_an_admin
+from functions.error_messages import *
```
IdezHD: ^^^

api/tests/test_user_roles.py
```diff
@@ -0,0 +1,162 @@
+from flask_bcrypt import Bcrypt
+
+from ..user_roles import *
```
IdezHD: See if you can remove this relational import, you should just be able to import 'user_roles'. ^^^

api/app.py
```diff
@@ -21,7 +21,7 @@
 app.config['SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
 app.config['SQLALCHEMY_COMMIT_ON_TEARDOWN'] = True
 app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
-app.config["JWT_SECRET_KEY"] = os.getenv('SUPER_SECRET_KEY')
+app.config["JWT_SECRET_KEY"] = str(os.getenv('SUPER_SECRET_KEY'))
```
IdezHD: the str should not be required. If you are wondering about the b in front of the string it tells python to turn it into a byte string. Returning this into a str could possibly break this

api/functions/update_user_role.py
```diff
@@ -2,6 +2,7 @@
...
 
 from flask_graphql_auth import *
```
IdezHD: ^^^

api/resolvers/users.py
```diff
@@ -1,7 +1,7 @@
...
+from user_roles import *
```
IdezHD: ^^^

api/app.py
peacheym: If you were to take a look at `/home/matt/.local/share/virtualenvs/api-q5ilRnRI/lib/python3.6/site-packages/jwt/utils.py` on line 24 there is a function called `force_bytes` that is called when any of the JWT algorithms run.  The JWT secret config has to be a string or else all tests that rely on the JWT package will fail. (SignIn is a big one)

peacheym:
Thanks for the comments @IdezHD.  All have been addressed.

PR #104, opened 2020-02-17
Models and DB update

IdezHD:
Updated scans table 'initiated_by' column to an integer that is a foreign key of a users id from the Users table


IdezHD:
This branch ended up conflicting with new changes that were merged into master and needs to be restarted due to versioning by flask-migrate

PR #105, opened 2020-02-17
Add robust tests to testing suite for user schema

peacheym:
This PR adds more tests against the user schema to ensure that all functionality works during every build of the `/api` directory .


api/cloudbuild.yaml
```diff
@@ -56,6 +56,7 @@ steps:
       - DB_PASS=$_DB_PASS
       - DB_PORT=$_DB_PORT
       - DB_USER=$_DB_NAME
+      - BASE32_SECRET=$_BASE32_SECRET
       - AZURE_DB_HOST=$_AZURE_DB_HOST
```
IdezHD: Just remove these azure things, so they don't get added back in

peacheym: Good catch!

PR #129, opened 2020-02-21
Users table username email fix

IdezHD:
- Removed user_email, username column from users table
- Renamed username to user_name
- Converted all user functions from user_email to user_name
- Converted all user queries and mutations from user_email to user_name
- Fixed all user tests to use the new user_name column


IdezHD:
Oh I also corrected each file I modified to the new formatting standard

PR #136, opened 2020-02-26
Add notification email endpoints to API

peacheym:
## Endpoints added
* sendPasswordReset
* sendValidationEmail

Both of these end points are able to return full details about the response being sent through the new 'NotificationEmail' object type.  This was useful for testing and also allows the front-end to know that it was sent successfully.

## Example usage
The following query is an example usage of the new endpoints that returns everything possible:

```
{
  sendPasswordReset(email: "<Valid Email>") {
    id
    reference
    uri
    scheduledFor

    content {
      body
      fromEmail
      subject
    }

    template {
      id
      uri
      version
    }
  }
}
```
And the simplest query to easily send the email:

```
{
  sendPasswordReset(email: "<Valid Email>") {
    id
}
```

## Todo
A small note: This PR contains code that must be updated to reflect the deployment URL's in each of the resolvers.


api/queries.py
```diff
@@ -1,65 +1,52 @@
...
+    resolve_generate_otp_url,
+)
+from scalars.url import URL
+from schemas.domains import Domains
+from schemas.groups import Groups
+from schemas.notification_email import (NotificationEmail)
+from schemas.organizations import Organizations
+from schemas.scans import Scans
+from schemas.sectors import Sectors
+from schemas.user import *
```
IdezHD: Should not be a star import, you could be importing unnecessary packages, that can cause errors. Also hard to tell what is being specifically used in file.

api/resolvers/notification_emails.py
```diff
@@ -0,0 +1,95 @@
...
+    NOTIFICATION_API_URL,
+)
+
+
+def resolve_send_password_reset(self, info, email):
+    template_id = password_reset_template()
+    password_reset_serial = URLSafeTimedSerializer(SUPER_SECRET_KEY)
+
+    # TODO: Change to deployment URL
+    password_reset_url = "http://localhost:3000/reset-password/" + password_reset_serial.dumps(
```
IdezHD: You should be able to get the url through flask requests. [Docs](https://flask.palletsprojects.com/en/1.1.x/api/#incoming-request-data)

api/resolvers/notification_emails.py
```diff
@@ -0,0 +1,95 @@
...
+    return populate_notification_email(response)
+
+
+def resolve_send_validation_email(self, info, email):
+    template_id = email_verification_template()
+
+    verify_email_serial = URLSafeTimedSerializer(SUPER_SECRET_KEY)
+
+    # TODO: Change to deployment URL
+    verify_email_url = "http://localhost:3000/verify-email/" + verify_email_serial.dumps(
```
IdezHD: ^^^

api/tests/test_notification_emails.py
```diff
@@ -0,0 +1,128 @@
...
+from functions.error_messages import scalar_error_type
+
+from functions.email_templates import (
+    email_verification_template,
+    password_reset_template)
+
+from manage import seed, remove_seed
+
+seed()
+from db import db
```
IdezHD: Check imports to ensure you're not importing things that are not needed

peacheym:
@IdezHD Take a look at my most recent changes!  I used `request.headers['Origin']` instead of `request.refferer` because I only needed server name and not entire path!

PR #137, opened 2020-02-26
User roles will never be the same

IdezHD:
User roles are now stored in the User_Affiliations table, this now can accommodate the edge cases of users having access to multiple organizations :tada:.

Changes:
- New database migration to accommodate new user role setup
- Admin and Admin_Affiliation tables have been removed, and all files related have also been removed
- New auth wrapper which requires a token sent through request headers, this wrapper also decodes the token and sends it into the query, or mutation. This functionality also includes checking that the token has not expired, and confirms the user claims and if there is a difference they will be directed to log back in.
- New auth functions, these work hand in hand with the return value of the auth wrapper. These allow you to customize individual queries and mutations with different level of access through out the function
- New function that takes a SQLAlchemy .all() query and converts to a list of dicts
- Sign In, Update User Role mutations have been updated following the new user role changes
- All user related tests have been updated to accommodate these changes
- Cloudbuild file now includes new salt env
- Many files have also been updated to meet new editor config formatting rules
- Cleaned up imports
- Removed unused files


api/functions/auth_functions.py
```diff
@@ -0,0 +1,71 @@
...
+    :param user_role: list of dicts containing user roles
+    :return: Returns true or false based on if this user is the given role
+    """
+    user_id = user_role[0]['user_id']
+    with app.app_context():
+        user_is_super_admin = User_affiliations.query \
+            .filter(User_affiliations.user_id == user_id) \
+            .filter(User_affiliations.permission == 'super_admin') \
+            .first()
+    if user_is_super_admin:
```
peacheym: Since this is a DB query, should there be a `is not None:` check?  Or is that implied with what you already have here?

api/functions/auth_wrappers.py
```diff
@@ -0,0 +1,79 @@
...
+    This function takes in a http requests and decodes the JWT sent by the user
+    :param request: The http request sent from the user
+    :return: Returns a list of dicts that contains the user claims
+    """
+    auth_header = request.headers.get('Authorization')
+    try:
+        payload = jwt.decode(auth_header, os.getenv('SUPER_SECRET_SALT'), algorithms=['HS256'])
+        return payload['roles']
+    except jwt.ExpiredSignatureError:
+        raise GraphQLError('Signature expired. Please login again')
```
peacheym: Errors as hard coded strings.  May be a non-issue, but I mentioned it in slack.

api/functions/generate_enums.py
```diff
@@ -2,93 +2,77 @@
...
-		with app.app_context():
-			query = db.session.query(Table).filter(getattr(Table, enum_column) == e.name).options(load_only(description_column))
-			row = db.session.execute(query).first()
+    description_dict = dict()
 
-			try:
-				description_dict.update({e.name: row[1]})
-			except:
-				description_dict.update({e.name: "Error loading description"})
+    for e in ret_enum:
```
peacheym: single letter variables like this are a dangerous game.  I assume it stands for 'enum' but it may be worth making it explicit.

api/functions/orm_to_dict.py
```diff
@@ -0,0 +1,10 @@
+def orm_to_dict(sql):
+    d1 = []
```
peacheym: Same idea here.  I really don't understand what `d1` and `d2` are.  I'd assume dict, but again I am a fan of being explicit

api/functions/update_user_role.py
```diff
@@ -1,36 +1,57 @@
...
-        user = User.query.filter(User.user_name == user_name)\
-                .update({'user_role': new_role})
-        db.session.commit()
-    else:
-        raise GraphQLError(error_not_an_admin())
+    if new_role == 'admin' or new_role == 'super_admin':
+        if is_super_admin(user_roles):
+            update_user_role_db()
+        else:
+            raise GraphQLError(str(error_not_an_admin()))
```
peacheym: I believe that the `str()` call here is redundant. IIRC, the function returns a string already wrapped in `str()`!

api/functions/update_user_role.py
```diff
@@ -1,36 +1,57 @@
...
-    if not user:
-        raise GraphQLError(error_role_not_updated())
-    else:
-        user = User.query.filter(User.user_name == user_name).first()
-        return user
+    elif new_role == 'user_read' or new_role == 'user_write':
+        if is_super_admin(user_roles) or is_admin(user_roles, org):
+            update_user_role_db()
+        else:
+            raise GraphQLError(str(error_not_an_admin()))
```
peacheym: ^^ Same with `str()` comment above

api/model_enums/roles.py
```diff
@@ -0,0 +1,24 @@
...
+            return 'A user who has been given access to run scans, and ' \
+                'manage domains'
+        elif self == RoleEnums.ADMIN:
+            return 'A user who has the same access as a user write account, ' \
+                   'but can define new user write accounts'
+        elif self == RoleEnums.SUPER_ADMIN:
+            return 'A user who has the same access as an admin, but can ' \
+                   'define new admins'
+        else:
+            return 'Another Role'
```
peacheym: This seems odd, can we return an error message instead of an ambiguous "other role"?? :)

api/resolvers/user_affiliations.py
```diff
@@ -0,0 +1,19 @@
...
+@require_token
+def resolve_test_user_claims(self, info, **kwargs):
+    """
+    This resolver returns the user_claims array -- A utility for testing
+    It requires that a JWT token be active, and that the user have an admin role
+    :returns: Returns the user_claims if user is an admin, raises error message if not.
+    """
+    roles = kwargs.get('user_roles')
+    if is_super_admin(roles):
+        return "Passed"
```
peacheym: What "Passed"?  The function?  The decorator?

api/tests/test_email_address_scalar.py
```diff
@@ -15,7 +15,8 @@
     os.path.realpath(os.path.join(os.getcwd(), os.path.expanduser(__file__))))
 sys.path.append(os.path.normpath(os.path.join(SCRIPT_DIR, PACKAGE_PARENT)))
 
-from scalars.email_address import EmailAddress, scalar_error_type, scalar_error_only_types
+from scalars.email_address import EmailAddress, scalar_error_type, \
```
peacheym: Should make this consistent with rest of project and use `from <package> import (<i1>, <i2>, ... , <in>)` form.  Just add the parentheses.

api/functions/auth_functions.py
IdezHD: It is implied, beautiful feature of python if statements

api/model_enums/roles.py
IdezHD: Unless we create another enum, and not create a description it will never appear

IdezHD:
Please hold off on doing reviews @tparrott-cse  and @sleepycat new work is being done for better tests

api/resolvers/user_affiliations.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    This resolver returns the user_claims array -- A utility for testing
+    It requires that a JWT token be active, and that the user have an admin role
+    :returns: Returns the user_claims if user is an admin, raises error message if not.
+    """
+    roles = kwargs.get('user_roles')
+    test_role = kwargs.get('role')
+    org = kwargs.get('org')
+
+    if test_role == 'super_admin':
```
peacheym: Would this long if / else / elif statement make more sense to be a switch statement?

IdezHD:
@sleepycat, @peacheym, and @tparrott-cse new tests have been written, and I believe in a much better way that fully tests the functionality of the auth functions and wrapper

api/resolvers/user_affiliations.py
IdezHD: It does make sense, sadly python dev team does not want us to have access to such beautiful things

IdezHD:
Something new as well auth functions now support higher level checks i.e. if you're an admin if there is a check for user write, you will pass the test, granted you have access to that org.

IdezHD:
Sorry guys, one more update to tests, will be pushed shortly

api/functions/auth_wrappers.py
```diff
@@ -0,0 +1,79 @@
...
+                user_roles.append(temp_dict)
+
+        user_claim_diff = list(
+            itertools.filterfalse(lambda x: x in user_claims, user_roles)) \
+                          + list(
+            itertools.filterfalse(lambda x: x in user_roles, user_claims))
+        if len(user_claim_diff):
+            print(user_claim_diff)
+            raise GraphQLError(
+                "User has a difference in their claims, Please sign in again.")
```
tparrott-cse: This won't make much sense to "normal" users, might be something specific we want to log but give the users something more generic.

PR #141, opened 2020-02-28
Initial depth check setup, testing has also been completed

IdezHD:
This PR will bring in a new security check that will not all users to submit a query of depth 10 or greater. The depth can be changed at any time inside the `api/backend/depth_check.py` file.

- Added backend directory to contain any backends we may choose to implement
- Added depth check backend to help mitigate cycle attacks
- Tests created to test depth check functionality

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -18,22 +18,21 @@ from queries import schema
 
 app = Flask(__name__)
 
-app.config[
-    'SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
+app.config['SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
 app.config['SQLALCHEMY_COMMIT_ON_TEARDOWN'] = True
 app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
 app.debug = True
 
 db.init_app(app)
 
-backend = DepthAnalysisBackend()
+backend = DepthAnalysisBackend(10)
 
 app.add_url_rule(
     '/graphql',
     view_func=GraphQLView.as_view(
         'graphql',
         schema=schema,
-        backend=DepthAnalysisBackend(),
+        backend=backend,
         graphiql=True
     )
 )
--- a/api/app.py
+++ b/api/app.py
@@ -4,7 +4,7 @@ from flask import Flask
 from flask_graphql import GraphQLView
 from waitress import serve
 
-from backend import SecurityAnalysisBackend
+from backend.security_check import SecurityAnalysisBackend
 
 from db import (
     db,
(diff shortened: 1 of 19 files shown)


api/app.py
```diff
@@ -23,11 +26,14 @@
...
 db.init_app(app)
 
+backend = DepthAnalysisBackend()
+
 app.add_url_rule(
     '/graphql',
     view_func=GraphQLView.as_view(
         'graphql',
         schema=schema,
+        backend=DepthAnalysisBackend(),
```
sleepycat: :heart_eyes:

api/backend/depth_check.py
```diff
@@ -0,0 +1,45 @@
...
+        document = super().document_from_string(schema, document_string)
+        ast = document.document_ast
+        for definition in ast.definitions:
+            # We are only interested in queries
+            if not isinstance(definition, FragmentDefinition):
+                if definition.operation != 'query':
+                    continue
+                if not isinstance(definition, FragmentSpread):
+                    depth = measure_depth(definition.selection_set)
+                    if depth > 10:  # set your depth max here
```
sleepycat: How about this:
```
backend=DepthAnalysisBackend({max_depth: 10})
```
Or maybe rename the whole thing:
```
backend=MaxDepth(10)
```
That might be clearer and easier to change.

api/tests/test_depth_check.py
```diff
@@ -0,0 +1,207 @@
...
+                          }
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+            }
+            ''', backend=backend)
+        result_refr = {
```
sleepycat: This isn't being used so we should get rid of it...

IdezHD: This is needed because the back end isn't apart of the schema so it has no way of knowing it is there. The back end is added in when we create the url rule

IdezHD: Sadly that is how flask-graphql is setup

api/app.py
IdezHD: :smile_cat:

IdezHD:
Also added in a default value of 10, this should make it easier to change and customize depending on usage.

IdezHD:
Hold off on review, I think I have something good to add in

IdezHD:
New Changes
- Added in cost checking
- Split depth check, and cost functions into their own files

api/app.py
```diff
@@ -15,19 +18,21 @@
...
-app.config[
-    'SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
+app.config['SQLALCHEMY_DATABASE_URI'] = f'postgresql+psycopg2://{DB_USER}:{DB_PASS}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
 app.config['SQLALCHEMY_COMMIT_ON_TEARDOWN'] = True
 app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = True
 app.debug = True
 
 db.init_app(app)
 
+backend = SecurityAnalysisBackend(max_depth=10, max_cost=1000)
```
sleepycat: :heart_eyes_cat:

PR #145, opened 2020-03-02
Add sectors and groups in a migration

sleepycat:
This commit creates the data for groups and sectors using a migration.
This is needed because the existing seeding doesn't provide all the values
needed, and isn't idempotent, meaning that it can't be run as an initContainer,
which gets run every time the pod starts.

### DIFF
--- a/api/migrations/versions/ffa0779887d6_.py
+++ b/api/migrations/versions/ffa0779887d6_.py
@@ -1,0 +1,165 @@
+"""empty message
+
+Revision ID: ffa0779887d6
+Revises: 0487ebd413ff
+Create Date: 2020-02-28 14:56:52.728247
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import table, column
+from sqlalchemy import String, Integer
+
+
+# revision identifiers, used by Alembic.
+revision = 'ffa0779887d6'
+down_revision = '0487ebd413ff'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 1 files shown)


api/migrations/versions/ffa0779887d6_.py
```diff
@@ -0,0 +1,165 @@
...
+			{id: 65, 'sector_id': 1, 's_group':'S4', 'description':'Maintain'},
+		]
+	)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # XXX: CASCADE nukes organisations here?
+    op.execute('truncate table groups CASCADE')
```
IdezHD: I would use `DELETE FROM groups` and sectors keeping tables intact

IdezHD:
Just realized this is a test breaking change, as we are creating the values per test which will conflict with data already in DB

api/migrations/versions/ffa0779887d6_.py
```diff
@@ -0,0 +1,165 @@
...
+		column('s_group', String),
+		column('sector_id', Integer),
+		column('description', String)
+	)
+
+
+    # TODO: so many duplicates!
+	op.bulk_insert(groups_table,
+		[
+			{id: 1, 'sector_id': 1, 's_group':'GC_A', 'description':'Arts'}, # duplicates sector
```
IdezHD: Check sector ID as they are all being assigned to the Sector with ID of 1

sleepycat:
Closing in favor of a different strategy.

PR #148, opened 2020-03-05
Add front end pages for CreateUser & SignInUser

peacheym:
# Additions
This PR adds two components namely `CreateUserPage` & `SignInPage`.  Both feature a `<Formik>` form, that onSubmit triggers a `useMutation()` react hook from the Apollo client that hits the GraphQL API appropriately.  Currently when the data is retrieved, nothing is done with it.  This is noted by To-do's in the code.

The functionality of processing the data shall be included in a next PR.  This is to reduce the size of the code review needed for these two front end components.


frontend/src/App.js
```diff
@@ -5,6 +5,8 @@ import { Global, css } from '@emotion/core'
 import { PageNotFound } from './PageNotFound'
 import { LandingPage } from './LandingPage'
 import { DomainsPage } from './DomainsPage'
+import {SignInPage} from "./components/SignInPage/SignInPage"
```
sleepycat: If we could get rid of the components directory that would be good. Essentially, everything is a component. I think we should stick with dogpiling them into the src directory for now and once we get this a little further advanced, and some good component boundaries we are happy with we can look to extract things into libraries.

frontend/src/App.js
```diff
@@ -5,6 +5,8 @@ import { Global, css } from '@emotion/core'
...
+import {CreateUserPage} from "./components/CreateUserPage/CreateUserPage"
```
sleepycat: I don't think your prettier is working.

frontend/src/components/CreateUserPage/CreateUserPage.js
```diff
@@ -0,0 +1,109 @@
...
+  Text,
+} from "@chakra-ui/core";
+import gql from 'graphql-tag'
+import {useMutation} from "@apollo/react-hooks";
+import {Link as RouteLink} from "react-router-dom";
+import {Field, Formik} from "formik";
+
+export function CreateUserPage(){
+    const [createUser, { loading, error, data }] = useMutation(gql`
+    mutation CreateUser($displayName: String!, $userName: EmailAddress!, $password: String!, $confirmPassword: String!) {
```
sleepycat: prettier++

PR #155, opened 2020-03-10
Graphql Schema Introduction

IdezHD:
This is quite a big shift for the API, the schema has been started from scratch (minus the user related stuff). Scans, and scan tags are not yet completed, but I would like to separate those into their own PR's rather than continuing in this one and create a massive PR.

- New Domain object built on brand new objects built on the different types of scans we have
- New resolvers for the domain object, including selection by url, organization, and if left blank will return all information that user has access to. This resolver does require that the user be logged in.
- New logger file has been started to try and mitigate the excess data that GraphQLErrors create
- Small auth bug fix


api/queries.py
```diff
@@ -57,154 +33,50 @@
...
-    get_scans_by_date_range = graphene.List(
-        of_type=Scans,
-        startDate=graphene.Argument(graphene.Date, required=True),
-        endDate=graphene.Argument(graphene.Date, required=True),
-        resolver=resolve_get_scans_by_date_range,
-        description="Allows selection of scans from a given date range"
-    )
-    get_scans_by_domain = graphene.List(
-        of_type=Scans,
+    users = SQLAlchemyConnectionField(UserObject._meta.connection, sort=None)
```
peacheym: `all_users` describes what this does more than `users`.  If the change was requested then its no problem!  Just my $0.02!

api/resolvers/domains.py
```diff
@@ -2,59 +2,148 @@
...
+            raise GraphQLError("Error, domain does not exist")
+        org_id = org_orm.id
+
+        # Check if user has read access or higher to the requested organization
+        if is_user_read(user_role, org_id):
+            query_rtn = query.filter(
+                Domains.domain == url
+            ).all()
+        else:
+            raise GraphQLError("Error, you do not have permission to view that domain")
```
peacheym: I think the `Error,` here is redundant?  Won't it say: Error: Error, you do not have permission.... ???

api/resolvers/domains.py
```diff
@@ -2,59 +2,148 @@
...
+    organization = kwargs.get('organization')
+    user_role = kwargs.get('user_roles')
+
+    # Generate list of org's the user has access to
+    org_id_list = []
+    for role in user_role:
+        org_id_list.append(role["org_id"])
+
+    if not org_id_list:
+        raise GraphQLError("Error, you have not been assigned to any organization")
```
peacheym: ^ Same here?  Disregard if above is incorrect!

api/resolvers/domains.py
```diff
@@ -2,59 +2,148 @@
...
+            else:
+                raise GraphQLError(
+                    "Error, you do not have permission to view that organization")
+
+        return query_rtn
+    else:
+        if is_super_admin(user_role=user_role):
+            query_rtn = query.all()
+            if not query_rtn:
+                raise GraphQLError("Error, no domains to view")
```
peacheym: ^^ Again, disregard for entire file if my recollection is incorrect.

api/schemas/README.md
```diff
@@ -0,0 +1,234 @@
...
+    ...
+
+
+schema = graphene.Schema(query=Query, mutation=Mutation)
+```
+To be able to access these new object types they have to be added to the schema.
+If the object is being implemented as a `Node` we want to use the `SQLAlchemyConnectionField`
+type with the `._meta.connection`. This will allow you to query this as a `relay.Node`. If you
+want to a single object type, it follows the same method as if you were adding it to another
+class via the `graphene.List(lambda: Object)` with a resolver querying the `Object.get_query(info)`.
```
peacheym: Bonus points for this file!  Thanks man!

api/schemas/domain/email_scan/spf/spf_tags.py
```diff
@@ -0,0 +1,44 @@
...
+from models import Spf_scans
+
+
+class SPFTags(SQLAlchemyObjectType):
+    """
+    Current settings of the currently configured SPF
+    """
+    class Meta:
+        model = Spf_scans
+        exclude_fields = (
```
peacheym: Short enough to be one line?  Or stack them 1 on each row! :)

api/schemas/domain/www_scan/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+
+class WWWScan(SQLAlchemyObjectType):
+    """
+    Results of HTTPS and SSL scans on domain
+    """
+    class Meta:
+        model = Scans
+        interfaces = (relay.Node,)
+        exclude_fields = (
+            "id", "domain_id",
```
peacheym: Same idea here, Either fit on one line, or one field per row!

api/schemas/domain/www_scan/https/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+from functions.get_timestamp import get_timestamp
+
+from schemas.domain.www_scan.https.https_tags import HTTPSTags
+
+
+class HTTPS(SQLAlchemyObjectType):
+    class Meta:
+        model = Https_scans
+        exclude_fields = (
+            "id", "https_scan"
```
peacheym: ^ This may seem nit-picky, But I do believe it will look cleaner!

api/schemas/domain/www_scan/ssl/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+from functions.get_domain import get_domain
+from functions.get_timestamp import get_timestamp
+
+
+class SSL(SQLAlchemyObjectType):
+    class Meta:
+        model = Ssl_scans
+        exclude_fields = (
+            "id", "ssl_scan"
```
peacheym: :)

api/queries.py
IdezHD: As it is selecting the users table I felt like it was a more appropriate name, I believe it will also be changed up in the future so it is more of a place holder for testing user functionality currently

api/resolvers/domains.py
IdezHD: No it will not have duplicated 'Error', it is however located in the `errors` list object

api/functions/auth_wrappers.py
```diff
@@ -73,5 +79,6 @@ def wrapper(self, *args, **kwargs):
             user_claims = check_user_claims(auth_resp)
             kwargs['user_roles'] = user_claims
             return method(self, *args, **kwargs)
-        raise GraphQLError(auth_resp)
+        # raise GraphQLError(auth_resp)
```
tparrott-cse: If not required, just delete please

PR #159, opened 2020-03-11
Sign in component redirects after success

peacheym:
# Additions
* Successful sign in redirects to `/`
* `/` path always refresh component
* ~~`localStorage()`~~ Apollo Client: `client.writeData()` used to hold jwt token
* tests added for these features.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -18,6 +18,8 @@ import { Footer } from './Footer'
 import { Navigation } from './Navigation'
 import { Flex, Link, CSSReset } from '@chakra-ui/core'
 import { SkipLink } from './SkipLink'
+import { useQuery, useApolloClient } from '@apollo/react-hooks'
+import gql from 'graphql-tag'
 
 export default function App() {
   const { i18n } = useLingui()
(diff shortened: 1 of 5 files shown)


frontend/src/App.js
```diff
@@ -45,29 +57,43 @@ export default function App() {
           <Link to="/domains">
             <Trans>Domains</Trans>
           </Link>
-          <Link to="/sign-in">
-            <Trans>Sign In</Trans>
-          </Link>
+
+          {// Dynamically decide if the link should be sign in or sign out.
+          window.localStorage.getItem('jwt') == null ? (
```
sleepycat: So, storing JWT's in localstorage is dicey. If someone can XSS this app, they can read the contents of localstorage.

That said, we are specifically avoiding the practices the make XSS so easy in so many apps: no remote scripts, [SRI hashes](https://www.srihash.org/) if we must, and using React... [which means we are building our UI from a tree of UI component instances](https://youtu.be/NcAYsC_TKCA?t=643) (each of which handle their input properly relative to the element that instance represents), instead of the ususal contatenating strings together and hoping the result is the HTML we expect.

If we're OK with asking users to log in again if they close their browser/tab, then we have another option for storing state: memory. Once the page loads, we can just hold stuff in memory either using [React Context](https://reactjs.org/docs/context.html#when-to-use-context) or Apollo Client's [local state management](https://www.apollographql.com/docs/react/data/local-state/). Let's explore those other options.

peacheym: Sounds good to me, Mike.  I will read up on those options until you're settled in and then we can chat. :smile:

peacheym: This issue is resolved as of my latest commit. I have went with the Apollo Client!

PR #177, opened 2020-03-20
Refactor SignInPage and remove warnings from tests

sleepycat:
This commit started as a refactor of the SignInPage, but I couldn't get the
tests to pass without warnings and some other refactoring and the result was
some mission creep.

The result is a far wider reaching refactor than I was planning but with some
nice results:
* A PasswordField component has been carved out of SignInPage
* SignInPage has been reworked and simplified and uses the new PasswordField
  component.
* SignInPage and related elements now have i18n
* All the tests pass without warnings


frontend/src/SignInPage.js
```diff
@@ -15,90 +18,93 @@ import {
...
-          tfaValidated
+  const { i18n } = useLingui()
+
+  const [signIn, { loading, error }] = useMutation(
+    gql`
+      mutation SignIn($userName: EmailAddress!, $password: String!) {
+        signIn(userName: $userName, password: $password) {
+          user {
+            userName
+            failedLoginAttempts
```
IdezHD: Hey Guys, was just taking a quick peak at this PR, and the user object has been updated and you can no longer access the `failedLoginAttempts` field for user

peacheym: Thanks @IdezHD,  I do not think it is actually being used for any functionality right now.  I will double check with Mike and let him update the PR! :smile:

frontend/src/__tests__/QRcodePage.test.js
```diff
@@ -14,7 +14,7 @@ import { ApolloProvider } from '@apollo/react-hooks'
 i18n.load('en', { en: {} })
 i18n.activate('en')
 
-describe('<SignInPage />', () => {
+describe('<QRcodePage />', () => {
```
peacheym: A classic mistake of mine,  thanks for the catch!

frontend/src/SignInPage.js
sleepycat: Ahh, nice catch. Lemme take care of that.

PR #180, opened 2020-03-20
Messing Around With Domains

IdezHD:
Domain mutations are finally here folks! :tada: 

- Creation, Modification, Removal of domains possible from user write level and higher
- Complete Testing for all mutations


IdezHD:
Found an issue, please hold off on further reviews

PR #198, opened 2020-03-23
Dmarc Report Page

peacheym:
# Additions
* Add Schema Faker
* Add React Pie Chart package
* Use GraphQL query for loading data and bind it to components.
* Add tests for Component rendering, Pass Icon, Fail Icon.


peacheym:
> Tests look great! Should MX be located on the DMARC page along with SPF (in lieu of DKIM)? MX records make up part of the DMARC scan results, for background

@Ethanljf I just spent a few moments scanning through the test data that was given to me and there are no distinct **mx** records in any of them.  Perhaps we can chat with the team and see where those come from, and if they need to be here at all!  Thanks for the review :D

IdezHD:
Before this gets merged into master, can you create a custom object for it, please and thank you!

peacheym:
> Before this gets merged into master, can you create a custom object for it, please and thank you!

Same comment as on the other PR @IdezHD

PR #199, opened 2020-03-24
Adding More Organizational Mutations

IdezHD:
Removing organizations have been brought to life in this PR :tada: 

- Mutation for removing organizations and all related information (Affiliations, Domains, Scans, etc.)
- Testing for all Organization Mutations
- Minor Bug Fixes

### DIFF
--- a/api/schemas/organizations_mutations.py
+++ b/api/schemas/organizations_mutations.py
@@ -68,6 +68,7 @@ class CreateOrganization(graphene.Mutation):
             acronym = cleanse_input(kwargs.get('acronym'))
             description = cleanse_input(kwargs.get('description'))
             zone = cleanse_input(kwargs.get('zone'))
+            sector = cleanse_input(kwargs.get('sector'))
             province = cleanse_input(kwargs.get('province'))
             city = cleanse_input(kwargs.get('city'))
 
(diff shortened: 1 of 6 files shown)


api/schemas/organizations_mutations.py
```diff
@@ -53,9 +65,10 @@ class Arguments:
         @require_token
         def mutate(self, info, **kwargs):
             user_id = kwargs.get('user_id')
-            acronym = cleanse_input(kwargs.get('acronym'))
+            acronym = cleanse_input(kwargs.get('acronym', ''))
```
peacheym: This seems weird, but I do not know enough about it.  Why do you have you have an empty string parameter for the `kwargs.get()` function???

IdezHD: This is to deal with the cleanse input functionality, if kwargs cant find the key in the dict then it returns none. The cleanse input function then has a fit about being passed a None type

peacheym: I think since `cleanse_input()` is a function that we wrote, it may be cleaner to handle this edge case in that function.  If you do not want to do it that way then I think a comment explaining the empty string would be necessary.

PR #200, opened 2020-03-24
Domain Mutation Time

IdezHD:
Domain Mutations Are Finally Here :tada: 

- Create, Update, and Remove Domains
- Testing for all mutation functionality, and access control

### DIFF
--- a/api/schemas/domains_mutations.py
+++ b/api/schemas/domains_mutations.py
@@ -5,7 +5,7 @@ from app import app
 from db import db
 
 from functions.auth_wrappers import require_token
-from functions.auth_functions import is_admin, is_super_admin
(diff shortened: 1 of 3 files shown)


api/schemas/domains_mutations.py
```diff
@@ -0,0 +1,234 @@
...
+
+            # Check to see if domain exists
+            domain_orm = db.session.query(Domains).filter(
+                Domains.domain == domain
+            ).first()
+
+            if domain_orm is not None:
+                raise GraphQLError("Error, Domain already exists.")
+
+            if is_super_admin(user_id=user_id) \
```
tparrott-cse: super admins should also pass is_user_write checks, so this is a bit redundant

PR #214, opened 2020-03-26
Bringing DMARC Reports to Everyone

IdezHD:
DMARC Report queries are finally here!

- Ability to retrieve DMARC reports through the domain/domains query.
- Ability to select reports based on a date range
- Started tests for domain/domains query returns (Will be completed once guidance tags are completed)


api/resolvers/dmarc_report.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    user_id = kwargs.get('user_id')
+    user_roles = kwargs.get('user_roles')
+    domain = cleanse_input(kwargs.get('domain'))
+    start_date = kwargs.get('start_date')
+    end_date = kwargs.get('end_date')
+
+    if (start_date is not None and end_date is None) or \
+        (start_date is None and end_date is not None):
+        raise GraphQLError('Error, both start and end dates are required.')
```
peacheym: The annotated file stated that `end_date` is far more important than `start_date`.  Do we think it is necessary to strictly enforce that both are required?

IdezHD: This is for selecting a date range, so I need two times to select between ?

peacheym: Ahh yes,  I see that now!  Looks good to me!! :smile:

PR #244, opened 2020-03-31
UserList Component

peacheym:
# Additions
This PR adds the UserList & PaginationButtons components.
* Adds two components
* Adds tests for both components (more for PBs)
* Adds onClick functions with place holder alerts to be incorporated in later PR.


frontend/schema.faker.graphql
```diff
@@ -2,7 +2,7 @@
 A field whose value is an upper case letter or an under score that has a
```
IdezHD, line 2: Can I get a custom object made for this please!

frontend/src/UserList.js
```diff
@@ -0,0 +1,210 @@
...
+
+import { Trans } from '@lingui/macro'
+import gql from 'graphql-tag'
+import { useQuery } from '@apollo/react-hooks'
+import { PaginationButtons } from './PaginationButtons'
+
+export function UserList() {
+  // This function generates the URL when the page loads
+  const { loading, error, data } = useQuery(
+    gql`
```
IdezHD, line 26: ^^^
With a custom object this can be sooooooooooo much cleaner

peacheym: @IdezHD I think it would be better to update the faker file in a separate pull request and then issue an additional PR to fix all of the data bindings in the UI.

PR #275, opened 2020-04-08
Manage user state with custom context and hook

sleepycat:
This commit adds a UserStateProvider function which makes available a
currentUser object and some helper functions to mutate that state.

The provider is passed an initial state object with the appropriate default
values for the currentUser object:

```javascript
<UserStateProvider initialState={{ userName: null, jwt: null, tfa: null }}>
  <App/>
</UserStateProvider>
```

That inital state is retrieved via the `useUserState()` hook along with helper
functions to operate on that state.

```javascript
const { currentUser, isLoggedIn, login, logout } = useUserState()
```

The `isLoggedIn` function will return true if there are truthy values in the
state (ie, we have non-null values for userName, jwt or tfa).

The `login` function expects the details for a user to be passed in, and will
set the currentUser accordingly:

```javascript
login({ userName: null, jwt: null, tfa: null })
```

The `logout` function works by setting the state back to whatever the initial
state was:
```javascript
logout()
```

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -16,29 +16,16 @@ import { Footer } from './Footer'
 import { Navigation } from './Navigation'
 import { Flex, Link, CSSReset } from '@chakra-ui/core'
 import { SkipLink } from './SkipLink'
-import { useQuery } from '@apollo/react-hooks'
-import gql from 'graphql-tag'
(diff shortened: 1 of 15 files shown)


peacheym:
Awesome stuff here @sleepycat ! Looking forward to using it :waffle:

PR #279, opened 2020-04-09
Create custom GraphQL types for UserPage and UserList

peacheym:
# Additions
* adds custom gql type for userPage
* adds custom gql type (and dependents) for userList
* Updates files and tests to use these new types.

### DIFF
--- a/frontend/schema.faker.graphql
+++ b/frontend/schema.faker.graphql
@@ -1109,21 +1109,24 @@ type UserList {
   """ Indicates which organization this list is being queried for."""
   organization: Acronym!
 
-  """ A list of userItems that will make up the list on the front end.  An item per user in the org."
-  userItems: [UserListItem]!
+  """Pagination data for this connection."""
+  pageInfo: PageInfo!
+
+  """Contains the nodes in this connection.  Aka: A list of userItems that will make up the list on the front end."""
+  edges: [UserItemEdge]!
 }
 
 type UserListItem implements Node {
     """The ID of the object."""
     id: ID!
 
-    """ The users email address or userName
+    """ The users email address or userName """
     userName: EmailAddress!
 
     """ The users display name"""
     displayName: String!
 
-    """ Indicates wether or not this user has enabled two factor authentication"
+    """ Indicates wether or not this user has enabled two factor authentication"""
     tfa: Boolean!
 
     """ Indicates if this user is an admin of the organization specified in UserList query."""
@@ -1134,7 +1137,7 @@ type UserListItem implements Node {
 This custom gql object is used to populate a userPage component in the front end.
 """
 type UserPage{
-    """ The users email address or userName
+    """ The users email address or userName"""
     userName: EmailAddress!
 
     """ The users display name"""
@@ -1146,9 +1149,18 @@ type UserPage{
     """ Indicates the preferred language of this user."""
     lang: String!
 
-    """ Indicates wether or not this user has enabled two factor authentication"
+    """ Indicates wether or not this user has enabled two factor authentication"""
     tfa: Boolean!
 
     """ Indicates if this user is an admin of the organization specified."""
     admin: Boolean!
+}
+
+"""A Relay edge containing a `UserItem` and its cursor."""
+type UserItemEdge {
+  """The item at the end of the edge"""
+  node: UserListItem!
+
+  """A cursor for use in pagination"""
+  cursor: String!
 }
--- a/frontend/schema.faker.graphql
+++ b/frontend/schema.faker.graphql
@@ -1121,16 +1121,16 @@ type UserListItem implements Node {
     id: ID!
 
     """ The users email address or userName """
-    userName: EmailAddress!
+    userName: EmailAddress! @fake(type: email)
 
     """ The users display name"""
-    displayName: String!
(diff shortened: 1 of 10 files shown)


frontend/schema.faker.graphql
```diff
@@ -1090,3 +1101,66 @@ type WWWScanEdge {
...
+"""
+type UserPage{
+    """ The users email address or userName"""
+    userName: EmailAddress! @fake(type: email)
+
+    """ The users display name"""
+    displayName: String! @fake(type: firstName)
+
+    """ Indicates which organization this users data is being displayed for."""
+    organization: Acronym! @examples(values: ["GC", "ABC", "ASDF", "NSTIR", "BC"])
```
IdezHD, line 1147: Just wondering if org and admin can be moved into its own type, and added as a list, just in case they have multiple affiliations and the admin seeing also has those affiliations, also they may be admin in multiple orgs

frontend/schema.faker.graphql
```diff
@@ -1090,3 +1101,66 @@ type WWWScanEdge {
...
+
+    """ Indicates the preferred language of this user."""
+    lang: String! @examples(values: ["English, "French])
+
+    """ Indicates wether or not this user has enabled two factor authentication"""
+    tfa: Boolean! @examples(values: [true, false])
+
+    """ Indicates if this user is an admin of the organization specified."""
```
IdezHD, line 1155: ^^^
There also isn't an org being specified here

frontend/src/__tests__/UserPage.test.js
```diff
@@ -0,0 +1,69 @@
...
+              <I18nProvider i18n={i18n}>
+                <UserPage />
+              </I18nProvider>
+            </ThemeProvider>
+          </MemoryRouter>
+        </MockedProvider>,
+      )
+    })
+  })
+})
```
sleepycat, line 69: If you asserted that the content you provided in the mocks was actually getting rendered this would be way better.
Since you are supplying the values in the mocks you can safely use `getByText` to find them in the output:
`const { getByText } = render(...)` and then `await waitFor(() => getByText('sometext'))`.

platform/overlays/gke/kustomization.yaml
```diff
@@ -10,7 +10,7 @@ images:
 - name: gcr.io/track-compliance/api
   newTag: master-4301b10
 - name: gcr.io/track-compliance/frontend
-  newTag: master-7ee5ba8
+  newTag: master-91c4299
```
sleepycat, line 13: This doesn't really belong in this commit.

frontend/src/__tests__/UserList.test.js
```diff
@@ -22,6 +34,256 @@ describe('<UserList />', () => {
...
+        <I18nProvider i18n={i18n}>
+          <MemoryRouter>
+            <MockedProvider mocks={mocks} addTypename={false}>
+              <UserList />
+            </MockedProvider>
+          </MemoryRouter>
+        </I18nProvider>
+      </ThemeProvider>,
+    )
+    expect(container).toBeDefined()
```
sleepycat, line 208: `container` [will always be defined](https://testing-library.com/docs/react-testing-library/api#container-1) here as it's the `<div>` within which the component is rendered.

frontend/src/UserList.js
```diff
@@ -99,112 +53,85 @@ export function UserList() {
...
-                    onClick={() => {
-                      window.alert('clicked box')
-                    }}
-                    _hover={{ borderColor: 'gray.200', bg: 'gray.200' }}
-                    p="30px"
+        ? data.userList.edges.map((edge) => {
+            return (
+              <PseudoBox
+                key={edge.node.id}
+                role="userCard"
```
sleepycat, line 61: The `role` attribute refers to an [ARIA role](https://www.w3.org/TR/using-aria/#aria-roles). There are fixed number of them defined in the standards and relied on by screen readers and other assistive technology. Most HTML elements have an implicit role (so a `<button />` has an implied `role=button`. 

PseudoBox here is probably a `<div>` by default, and the non-standard role of `userCard` is being added to it can be found by [Testing Library's `getAllByRole` function](https://testing-library.com/docs/dom-testing-library/api-queries#byrole) later. I don't know if this would have actively negative consequences for assitive tech users (given that that role isn't a real ARIA role), but it's not a good idea.

Breaking these components down smaller and [favouring composition](https://youtu.be/3XaXKiXtNjw?t=832) makes it easier to test. Smaller components can [spread props directly onto certain elements](https://github.com/canada-ca/tracker/blob/c66987d43e5bbe8c5a18931e2470f8a8ddc74865/frontend/src/EmailField.js#L27) they render, [allowing the use of `getByTestId`](https://github.com/canada-ca/tracker/blob/c66987d43e5bbe8c5a18931e2470f8a8ddc74865/frontend/src/__tests__/EmailField.test.js#L29) for really simple selections. This would be a better strategy to the way roles are being used here.

peacheym: @sleepycat Mike, thanks for the awesome feedback.  Do you think a "userCard component would be better here???

peacheym:
@sleepycat please see my latest commit where I attempt to resolve all of the issues your found at the end of last week.  Would love to chat about it more! :smile:

PR #400, opened 2020-05-12
Here Comes The Slug

nslandolt:
Finally the conversion over to slugs has been made.

- Created new slug scalar
- Converted all instances of org acronyms to org slugs
- Added slugs to domains ( a slugified version of their domain)
- Fixed all tests to match this new tech stack
- Moved `session.commit()` outside the loop so it is done once all the deletes have been done reducing the amount of db transactions


api/migrations/versions/c264f0905a36_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'c264f0905a36'
+down_revision = 'b781fb62a115'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('domains', sa.Column('slug', sa.String(), nullable=True))
+    op.create_index(op.f('ix_domains_slug'), 'domains', ['slug'], unique=False)
```
sleepycat, line 22: :brain:

PR #435, opened 2020-05-21
Verify Your Account Today!

nslandolt:
Email verification on user sign-up is a go :rocket: but if that fails there is a backup mutation that will allow users to request another email be sent. 

- New `email_validated` field on the Users model, and new hybrid method `verify_account`
- Functionality added to `signUp` mutation that will send a verification email
- `sendEmailVerification` mutation allows the users to request another verification email be sent
- `emailVerifyAccount` mutation allows users to verify their account through a token link sent to their inbox
- User sandbox org is created once the user has verified their account now
- Tests for all new functionality included
- Tests corrected for new `verify_account` when querying their own sandbox org
- `frontend/schema.faker.graphql` has been updated to reflect new changes
- Minor changes and fixes


api/functions/verification_email.py
```diff
@@ -0,0 +1,61 @@
...
+    This function allows a user object to be passed in during account creation
+    and send an email to be used for verifying accounts
+    :param user: A instance of /models/User.py
+    :return: None
+    """
+    # Create Notify Client
+    notify_client = NotificationsAPIClient(
+        api_key=os.getenv("NOTIFICATION_API_KEY"),
+        base_url=os.getenv("NOTIFICATION_API_URL"),
+    )
```
sleepycat, line 23: I think this is a dependency that should be passed in rather than instantiated here. This would make the `send_verification_email` function way easier to test.
Maybe something like `def send_verification_email(user, client = NotificationsAPIClient(...)):`. 
That way in the tests you coud pass in a [stub](https://github.com/alex/pretend#how-do-i-use-pretend) and assert that the `send_email_notification` method is called on it with the args you expected. [MagicMock](https://docs.python.org/dev/library/unittest.mock.html) is another way to create an object that will let you do this.
This make sure our tests aren't hitting the network and Notify when they run.

api/models/Users.py
```diff
@@ -73,3 +60,17 @@ def password(self, password):
...
+        self.email_validated = True
+
+        # Create users sandbox org
+        acronym = slugify_value(self.user_name).upper()[:50]
+        self.user_affiliation.append(
+            User_affiliations(
+                permission="admin",
+                user_organization=Organizations(name=self.user_name, acronym=acronym,),
+            )
+        )
```
sleepycat, line 76: :star_struck: Awesome seeing this stuff moved into the model!

api/functions/verification_email.py
```diff
@@ -0,0 +1,61 @@
...
+        "status"
+    )
+
+    if (
+        email_status == "permanent-failure"
+        or email_status == "temporary-failure"
+        or email_status == "technical-failure"
+    ):
+        return "Email Send Error: {}".format(email_status)
+    else:
```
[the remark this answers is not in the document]
nslandolt: So with a bit of testing Notify does not send back any response code unless its an error code. The only way to get the status of the message is to actually request it after the fact using the functionality just below.

api/functions/verification_email.py
nslandolt: So for testing the actually functionality we have been using our test key, and following along with the Notify [docs](https://docs.notifications.service.gov.uk/python.html#other-testing) in regards to testing.

api/functions/verification_email.py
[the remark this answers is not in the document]
sleepycat: Fine. Let's do this for now and figure out something better soon.

api/functions/verification_email.py
sleepycat: I looked and the thing I was talking about needs a few other changes. Let's get this in, and refactor it after.

api/db.py
```diff
@@ -9,10 +9,11 @@
 DB_PORT = os.getenv("DB_PORT")
 DB_NAME = os.getenv("DB_NAME")
 
-
+# Add this in later branch -> pool_pre_ping=True,
```
tparrott-cse, line 12: why not now?

PR #438, opened 2020-05-25
Scanner code reworking

Ethanljf:
Implementation of ASGI within scanner services as well as general refactoring. Support for testing with the use of mock data.


scanners/dispatcher/dispatcher.py
```diff
@@ -1,179 +1,108 @@
-import os
+import ast
+import emoji
 import sys
 import logging
 import requests
-import jwt
-import threading
-from flask import Flask, request
+from starlette.applications import Starlette
```
nslandolt, line 6: This is all I want for Christmas, but you know who says no :sob:

scanners/dispatcher/dispatcher.py
```diff
@@ -1,179 +1,108 @@
...
-
-@app.route("/receive", methods=["GET", "POST"])
-def receive():
-
-    payload = {}
-    dkim_flag = False
+def initiate(received_payload, scan_type):
 
     try:
-        decoded_payload = jwt.decode(
```
nslandolt, line 45: Is the JWT gone forever, if so do you want to update the api to reflect the changes, or do you want me to ?

scanners/dmarc/dmarc_scanner.py
```diff
@@ -107,16 +64,44 @@ def scan(scan_id, domain):
...
         logging.error(
-            "(SCAN: %s) - Failed to check the given domains for DMARC/SPF records: %s"
-            % (scan_id, e)
+            "Failed to check the given domains for DMARC/SPF records: %s"
+            % str(e)
         )
         return None
 
-    if scan_result["dmarc"]["record"] is "null":
+    if scan_result["dmarc"]["record"] == "null":
```
nslandolt, line 72: Just wondering if this value is always available to retrieve from, may be a good idea to switch to `.get()` just in case it cannot be found.

Ethanljf: Good catch. Updated that in the event it isnt.

scanners/dispatcher/dispatcher.py
Ethanljf: Updated

scanners/dispatcher/dispatcher.py
Ethanljf: 😂

scanners/dispatcher/requirements.txt
```diff
@@ -1,4 +1,4 @@
-Flask
-waitress
-PyJWT
+starlette
+uvicorn
+emoji
```
sleepycat, line 3: :heart:

PR #446, opened 2020-05-25
Update Sign-in and Create User Mutations

k-sullivan:
Changed front end Sign-in and Create User Mutations.

- Add show/hide button to PasswordConfirmation component
- Add language selection for user creation
- Sign out actually works
- Auto sign in on account creation


frontend/src/CreateUserPage.js
```diff
@@ -4,56 +4,85 @@ import { useMutation } from '@apollo/react-hooks'
...
+        {i18n._(t`Create an account by entering an email and password.`)}
       </Text>
       <Formik
         validationSchema={validationSchema}
         initialValues={{ email: '', password: '', confirmPassword: '' }}
-        onSubmit={async (values) => {
+        onSubmit={async (values, _actions) => {
           console.log('values', values)
-          createUser({
+          await signUp({
```
sleepycat, line 79: You don't need for the `await` keyword here, you're passing an async function to Formik, which returns a promise when executed. Formik uses the promise and it's eventual resolution to set the [isSubmitting](https://jaredpalmer.com/formik/docs/api/formik#onsubmit-values-values-formikbag-formikbag--void--promiseany) value automatically.

frontend/src/CreateUserPage.js
```diff
@@ -4,56 +4,85 @@ import { useMutation } from '@apollo/react-hooks'
...
     <Stack spacing={2} mx="auto">
       <Text mb={4} fontSize="2xl">
-        Create an account by entering an email and password.
+        {i18n._(t`Create an account by entering an email and password.`)}
       </Text>
       <Formik
         validationSchema={validationSchema}
         initialValues={{ email: '', password: '', confirmPassword: '' }}
-        onSubmit={async (values) => {
+        onSubmit={async (values, _actions) => {
```
sleepycat, line 77: If you aren't used this variable it can just be removed. Save the leading underscore for scenarios where like skipping the first argument.

frontend/src/SignInPage.js
```diff
@@ -84,7 +82,7 @@ export default function SignInPage() {
         validationSchema={validationSchema}
         initialValues={{ email: '', password: '' }}
         onSubmit={async (values, _actions) => {
-          await signIn({
+          await authenticate({
```
sleepycat, line 85: Same as above, remove unused variable and drop the extra await.

frontend/src/UserPage.js
```diff
@@ -181,7 +167,7 @@ export default function UserPage() {
           w={'50%'}
           onClick={() => {
             // This clears the JWT, essentially logging the user out in one go
-            client.writeData({ data: { jwt: null, tfa: null, userName: null } }) // How is this done?
+            logout()
```
sleepycat, line 170: Nice!

PR #453, opened 2020-05-27
Changes required for #446, clean up credential components

k-sullivan:
Adds changes for #446 (whoops). Also added:

- PasswordConfirmation validation done with Yup schema in parent component
- Validation for LanguageSelect
- Removed extra validation check for credential components (isRequired)
- Cleaned up/ added translations
- Margins/attributes for EmailField, LanguageSelect, Password, and PasswordConfirmation now passed as props


frontend/src/PasswordField.js
```diff
@@ -13,25 +14,25 @@ import {
...
       <InputGroup size="md">
         <InputLeftElement>
           <Icon name="lock" color="gray.300" />
         </InputLeftElement>
 
         <Input
-          {...props}
           {...field}
-          isInvalid={meta.error}
+          data-testid={dataTestId !== 'undefined' && dataTestId}
```
sleepycat, line 32: Is there a reason why we are [spreading the PasswordFields props](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes) onto the FormControl component instead of the the Input?
The reason to spread those props onto the input was that the input is really the point of the component anyway and by doing so you can add attributes to it like `<PasswordField dataTestId="pwfield" name="password" />` in the tests and have them show up on the element we expect.
The `dataTestId` thing is a way of [finding elements in a test](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid) that [Testing Library](https://testing-library.com/) provides. We really shouldn't have that showing up in our production code.

If you switch it back to `<FormControl isInvalid={meta.error && meta.touched}>` and spread the props on the Input component instead with `<Input {...props} />` you can delete that dataTestId thing (and remove it from the propTypes too) and the tests should still work fine.

PR #535, opened 2020-06-12
Pgo cluster a go go

tparrott-cse:
This pull request replaces the PostgreSQL container used up until now with a Postgres Operator.  By making this change, we are enabling HA data features in our deployment.


platform/overlays/gke/istio.yaml
```diff
@@ -6731,12 +6731,18 @@ metadata:
 spec:
   loadBalancerIP: 34.95.5.243
   ports:
+  - name: status-port
+    port: 15021
+    targetPort: 15021
```
sleepycat, line 6736: These three lines and three for port 15443 should be removed. This exposes health check data and shows up in port scans which is a little gross. I don't have an automated way of removing this yet, but I'll figure out something soon.

PR #576, opened 2020-06-23
Implementation of automated scanning service

Ethanljf:
Changes include the implementation of an automated scanning service (and corresponding tests), which enqueues domains for scanning by gleening domains from the database at a specified interval. All retrieved domains are passed to the dispatcher.


services/auto-scan/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends apt-utils && \
+apt-get install -y --no-install-recommends python3 python3-pip && \
+apt-get install -y --no-install-recommends python3-setuptools python3-wheel
+
+# Install dependencies.
+RUN python3 -m pip install -r requirements.txt
+
+ENTRYPOINT ["python3", "autoscan.py"]
```
sleepycat, line 16: If we don't create/specify a user this will end up running as root. Unless someone does something like [userns-remap](https://docs.docker.com/engine/security/userns-remap/) root in the container is going to be root on the host. There is a good example of creating a user [here](https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf).

services/auto-scan/cloudbuild.yaml
```diff
@@ -0,0 +1,48 @@
...
+    args: ["-c", "docker run --net cloudbuild --name testdb -d --rm -p 127.0.0.1:5432:5432 -e POSTGRES_USER=track_dmarc -e POSTGRES_PASSWORD=postgres postgres:alpine"]
+
+  - name: mikewilliamson/wait-for
+    id: wait_testdb
+    args: ['testdb:5432']
+
+  - name: 'gcr.io/cloud-builders/docker'
+    id: start_dbmigrate
+    entrypoint: /bin/sh
+    args: ["-c", "docker run --net cloudbuild --name dbmigrate -d --rm -e POSTGRES_USER=track_dmarc -e POSTGRES_PASSWORD=postgres -e DB_HOST=testdb gcr.io/track-compliance/services/db-migration"]
```
sleepycat, line 14: I think this would do the same thing: 
```
- name: "gcr.io/track-compliance/services/db-migration"
    id: start_dbmigrate
    env: 
    - POSTGRES_USER=track_dmarc
    - POSTGRES_PASSWORD=postgres
    - DB_HOST=testdb
```

PR #577, opened 2020-06-24
Make URL type more strict

sleepycat:
The existing URL type allowed any valid URL, but we don't actually want *all*
URLs. Urls with user credentials in them, mongodb, web sockets, file,
javascript or telephone protocols... there are a bunch of things that are
technically valid but that we really don't want people passing us because they
are dangerous, useless, or we shouldn't/won't/can't scan it.

This is an attempt to tighten our definition of what constitutes a valid url for
us.

### DIFF
--- a/api/scalars/url.py
+++ b/api/scalars/url.py
@@ -1,13 +1,17 @@
+from urllib.parse import urlparse
 from re import compile
+
 from graphene.types import Scalar
 from graphql.language import ast
 from graphql import GraphQLError
 
+# >>> urlparse("https://asdf.com")
+# ParseResult(scheme='https', netloc='asdf.com', path='', params='', query='', fragment='')
+
+
 from functions.error_messages import *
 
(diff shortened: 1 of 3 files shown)


api/scalars/url.py
```diff
@@ -1,13 +1,17 @@
...
-from functions.error_messages import *
+# >>> urlparse("https://asdf.com")
+# ParseResult(scheme='https', netloc='asdf.com', path='', params='', query='', fragment='')
+
 
-URL_REGEX = r"[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)"
+from functions.error_messages import *
 
-URL_REGEX_CHECK = compile(URL_REGEX)
+safe_schemes = ["https", "http", "mailto", "ftp"]
```
nslandolt, line 14: As far as I have seen with the test data, and what's coming from the cosmos db, none of the domains that are going to be used in the service do not actually have a schema such as the above actually attached to them, it will just be the raw domain. Looking at the packages we are using for the scanning as well it seems that they do not require a schema to be included.

PR #672, opened 2020-07-10
People can finally be added to organizations

nslandolt:
A new query that allows admins of organizations to invite users to the organization. If the user the admin wants to invite does not have an account however, they will receive an email containing a link to create an account with a token in the url, that upon user creation will automatically assign that new user account to that organization.

### DIFF
--- a/api/tests/test_invite_user_to_org.py
+++ b/api/tests/test_invite_user_to_org.py
@@ -1,0 +1,261 @@
+import logging
+import pytest
+import pytest_mock
+
+from pytest import fail
+
+from db import DB
+from models import Organizations, Users, User_affiliations
+from json_web_token import tokenize
+from tests.test_functions import json, run
+
+
+@pytest.fixture()
+def db():
+    s, cleanup, db_session = DB()
+    yield s, db_session
+    cleanup()
+
+
+def test_successful_invite_existing_user_to_org_existing_user(caplog, db, mocker):
+    """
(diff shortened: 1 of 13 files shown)


api/tests/test_invite_user_to_org.py
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+
+    affiliation = (
+        session.query(User_affiliations)
+        .filter(User_affiliations.user_id == user.id)
+        .filter(User_affiliations.organization_id == org_one.id)
+        .filter(User_affiliations.permission == "user_read")
+        .first()
+    )
```
sleepycat, line 100: I think this could also be written as:
```
    affiliation = (
        session.query(User_affiliations)
        .filter_by(user_id = user.id, organization_id = org_one.id, permission = "user_read")
        .first()
    )
```
Maybe cleaner?

frontend/src/graphql/mutations.js
```diff
@@ -9,11 +9,13 @@ export const SIGN_UP = gql`
...
-      password: $password
-      confirmPassword: $confirmPassword
-      preferredLang: $preferredLang
+      input: {
+        displayName: $displayName
+        userName: $userName
+        password: $password
+        confirmPassword: $confirmPassword
+        preferredLang: $preferredLang
+      }
```
sleepycat, line 18: :heart_eyes:

PR #708, opened 2020-07-21
Update input_validators.py

peacheym:
A very fast typo fix :smile:


peacheym:
> Thank you for the fix, point_down is how I type
> 
> 

Honestly, it may have been me in January.

PR #811, opened 2020-09-09
Make sign up invite only

nslandolt:
This change to the sign-up mutation will only allow users to sign up if they were sent an invite email. We will use the initial SA account to send out the initial invites and go from there.


Ethanljf (approved):

sleepycat (approved):
LGTM!

PR #835, opened 2020-09-20
External Reporting Check

Ethanljf:
These changes introduce a follow-up check to ensure that external DMARC report recipients found within the record's 'RUA' tag are indeed configured to receive/accept reports. Such configuration is a strong indicator of an authorized external reporting arrangement between the original organizational domain and the recipient.

More info can be found [here](https://tools.ietf.org/html/rfc7489#section-7.1)

### DIFF
--- a/services/scanners/dns/dns_scanner.py
+++ b/services/scanners/dns/dns_scanner.py
@@ -51,10 +51,11 @@ async def scan_dmarc(domain):
 
     # Retrieve 'rua' tag address.
     try:
-        rua_addr = scan_result['dmarc']['tags']['rua']['value'][0]['address']
+        rua_addr = scan_result["dmarc"]["tags"]["rua"]["value"][0]["address"]
     except KeyError as e:
         logging.error(
-            "Failed to retrieve RUA address from tag value. Report acceptance check will not be performed. (KeyError: %s)" % str(e)
+            "Failed to retrieve RUA address from tag value. Report acceptance check will not be performed. (KeyError: %s)"
+            % str(e)
         )
         rua_addr = None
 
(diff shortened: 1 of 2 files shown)


services/scanners/dns/dns_scanner.py
```diff
@@ -34,19 +35,49 @@ async def scan_dmarc(domain):
...
+        )
+        rua_addr = None
+
+    if rua_addr is not None:
+        # Extract the domain from the address string (e.g. 'dmarc@cyber.gc.ca' -> 'cyber.gc.ca').
+        rua_domain = rua_addr.split("@", 1)[1]
+
+        # If the report destination differs from the originating domain, assert reports are being accepted.
+        if rua_domain == domain:
+            scan_result['dmarc']['tags']['rua']['accepting'] = True
```
michaeldavie-cyber, line 66: If `rua_domain` has the same organizational domain as `domain` (i.e. one level below the public suffix), we don't need to do the check. So, `tracker.cyber.gc.ca` sending to `cyber.gc.ca` should pass, since `gc.ca` is a public suffix.

There's a public suffix library at https://pypi.org/project/publicsuffixlist/

tparrott-cse (approved):

michaeldavie-cyber (changes requested):
Made a comment inline

services/scanners/dns/dns_scanner.py
```diff
@@ -61,8 +62,14 @@ async def scan_dmarc(domain):
         # Extract the domain from the address string (e.g. 'dmarc@cyber.gc.ca' -> 'cyber.gc.ca').
         rua_domain = rua_addr.split("@", 1)[1]
 
-        # If the report destination differs from the originating domain, assert reports are being accepted.
-        if rua_domain == domain:
+        # Extract organizational domain from original domain (e.g. 'tracker.cyber.gc.ca' -> 'cyber.gc.ca')
+        extract = tldextract.TLDExtract(include_psl_private_domains=True)
+        extract.update()
+        parsed_domain = extract(domain)
+        org_domain = parsed_domain.domain + parsed_domain.suffix
```
michaeldavie-cyber, line 69: `'.'join([parsed_domain.domain, parsed_domain.suffix])`

I think we need to extract the organizational domains of both and compare them.

michaeldavie-cyber (changes requested):

michaeldavie-cyber:
Looks good!

PR #843, opened 2020-09-22
Forbid -> Replace

Ethanljf:
Contrary to prior understanding, a pod concurrency policy of "Forbid" results in blocking behaviour. Even after a scheduled scan has been performed, it will be unable to switch status to "Completed" to avoid the generation of concurrent 'autoscan' pods. 

This has been updated to reflect a policy of 'Replace', which enables the desired behaviour.


tparrott-cse (approved):

sleepycat (approved):
:+1:

PR #847, opened 2020-09-23
Istio 1.7 update

Ethanljf:
These changes bring us to Istio 1.7!

Notable changes include an overhaul of the way addon components work. Thus, rather than generating istio.yaml with Kiali/Jaeger enabled, they have been added into `platform/overlays/gke` as individual manifests. 

[More info here](https://istio.io/latest/news/releases/1.7.x/announcing-1.7/change-notes/)


sleepycat (approved):
:+1:

PR #878, opened 2020-10-02
Implementation of 'Find my domains' query and underlying loader

Ethanljf:
These changes introduce the find-my-domains query and load-domain-connections-by-user-id loader to the in-progress JavaScript API.

### DIFF
--- a/api-js/src/loaders/domains/load-domain-connections-by-user-id.js
+++ b/api-js/src/loaders/domains/load-domain-connections-by-user-id.js
@@ -1,0 +1,164 @@
+const { aql } = require('arangojs')
+const { fromGlobalId, toGlobalId } = require('graphql-relay')
+
+const domainLoaderConnectionsByUserId = (
+  query,
+  userId,
+  cleanseInput,
+) => async ({ after, before, first, last }) => {
+  let afterTemplate = aql``
+  let beforeTemplate = aql``
+
+  const userDBId = `users/${userId}`
+
+  if (typeof after !== 'undefined') {
+    const { id: afterId } = fromGlobalId(cleanseInput(after))
+    afterTemplate = aql`FILTER TO_NUMBER(domain._key) > TO_NUMBER(${afterId})`
+  }
+
+  if (typeof before !== 'undefined') {
+    const { id: beforeId } = fromGlobalId(cleanseInput(before))
+    beforeTemplate = aql`FILTER TO_NUMBER(domain._key) < TO_NUMBER(${beforeId})`
+  }
+
+  let limitTemplate = aql``
+  if (typeof first !== 'undefined' && typeof last === 'undefined') {
+    limitTemplate = aql`SORT domain._key ASC LIMIT TO_NUMBER(${first + 1})`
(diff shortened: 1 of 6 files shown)


api-js/src/loaders/domains/load-domain-connections-by-user-id.js
```diff
@@ -0,0 +1,164 @@
...
+  } else if (typeof first !== 'undefined' && typeof last !== 'undefined') {
+    console.warn(
+      `User: ${userId} tried to have first and last set in domain connection query`,
+    )
+    throw new Error(
+      'Error, unable to have first, and last set at the same time.',
+    )
+  }
+
+  let affiliatedOrgsCursor
```
nslandolt, line 38: So I have come up with this new way of querying that reduces the chances of a db, cursor failure and that is to do it all in one step
```
dkimScansCursor = await query`
LET dkimIds = (FOR v, e IN 1 ANY ${domainId} domainsDKIM RETURN e._to)
FOR dkimScan IN dkim
    FILTER dkimScan._id IN dkimIds
    ${afterTemplate}
    ${beforeTemplate}
    ${limitTemplate}
    RETURN dkimScan
`
```

In your case it would look something like
```
LET userAffiliations = (FOR v, e IN 1..1 ANY ${userDBId} affiliations RETURN e._from)
FOR userAffiliation IN userAffilitations
    LET orgClaims = (FOR v, e IN 1..1 ANY userAffiliation claims RETURN e._to)
    FOR domain IN domains
      FILTER orgClaims[** FILTER CURRENT == domain._id]
      ${afterTemplate}
      ${beforeTemplate}
      ${limitTemplate}
      RETURN domain
```

api-js/src/loaders/domains/load-domain-connections-by-user-id.js
```diff
@@ -0,0 +1,164 @@
...
+    console.error(
+      `Cursor error occurred while user: ${userId} was trying to gather affiliated organizations in loadDomainsByUser.`,
+    )
+    throw new Error('Unable to load organizations. Please try again.')
+  }
+
+  let claimedDomainsCursor, claimedDomains, orgId
+  const orgDomains = []
+
+  for (let i = 0; i < affiliatedOrgs.length; i++) {
```
nslandolt, line 64: Look at the above comment, this way we only ever have to send one query, so it won't be prone to failing as often when looping through Id's

api-js/src/queries/domains/find-my-domains.js
```diff
@@ -7,7 +7,31 @@ const findMyDomains = {
...
+      userId,
+      loaders: {
+        domainLoaderConnectionsByUserId,
+      },
+    },
+  ) => {
+    let domainConnections
+
+    try {
+      domainConnections = await domainLoaderConnectionsByUserId({args})
```
nslandolt, line 23: Unless you use the spread operator, you don't need to include the curly braces because then you will be wrapping an object inside and object, and the function won't be able to de-structure it properly

nslandolt (changes requested):

nslandolt (approved):

PR #938, opened 2020-10-11
Find my organizations

Ethanljf:
Implementation of find-my-organizations query and underlying load-organization-connections-by-user-id + all tests


nslandolt (approved):
Amazing as alwayd

PR #939, opened 2020-10-11
Implementation of find-organization-by-slug query

Ethanljf:
Adds org slug querying to the JS API. Tests included + ran prettier


nslandolt (approved):
Once again very good

PR #948, opened 2020-10-13
find-domains-by-org

Ethanljf:
Implementation of find-domains-by-org and associated tests.


api-js/src/queries/domains/find-domains-by-org.js
```diff
@@ -14,7 +14,60 @@ const findDomainsByOrg = {
     },
     ...connectionArgs,
   },
```
nslandolt, line 16: You can add a `deprecated:true` here I think to inform the front end guys that there's a different way to do this

nslandolt (changes requested):
Looking good, again I am sorry

PR #1115, opened 2020-11-03
User affiliation base type

Ethanljf:
Implementing user affiliation base type and corresponding fields in user and organization objects.


nsdeschenes (approved):
:ship: it

PR #1131, opened 2020-11-05
Add CI for the JS API

sleepycat:
So there is a surprising amount of stuff going on here, but it breaks down in
to 3 parts.

First, dotenv is now required in the src/setupEnv.js (and the jest.config.js
has been updated accordingly) instead of being in every test file, and it pulls
in a file you can put test appropriate values in: `test.env`. Vestigial config
from package.json has been removed since it's overwritten by jest.config.js (to
my chagrin).

Second, A Cloudbuild trigger has been created to use the updated cloudbuild.yaml.

Third, updates the to README to explain.


nsdeschenes (approved):
🚢 it

PR #1210, opened 2020-11-18
Super Admin Service

nsdeschenes:
This service is to provide an `initContainer` to be ran before the API starts up to ensure that a super admin account, org, and the required affiliation are present in arango.

- Process
  1. Checks to see if account or org is present
  2. Depending on findings the service will take appropriate actions to remedy the problem
      - Will create account if not found
      - Will create org if not found
      - Removes old affiliation if org or account are found, and will create a new affiliation
  3. Program will exit when completed
- Env based values
  - All account, and org details are determined by env's so they are customizable to meet our requirements
- Testing
  - 100% coverage across the board


Ethanljf (approved):

sleepycat (approved):
:heart_eyes_cat:

tparrott-cse (approved):

PR #1302, opened 2020-12-04
Refactor To CheckDomainPermission Function

nsdeschenes:
This is a small refactor, adding in a check to see if the user has super admin permission, if so it bypasses the old check.


tparrott-cse (approved):
YUUUUUUUP

PR #1423, opened 2021-01-07
Invite User To Org Update

nsdeschenes:
This is a quick update to fix the routing issue in the email sent to the invited user.


tparrott-cse (approved):
I like ones like this :)

PR #1432, opened 2021-01-11
Perform scans at 1am every second day

Ethanljf:
Like the title says :)


tparrott-cse (approved):
Make it shorter next time please

PR #1454, opened 2021-01-14
Remove affiliations field from signUp mutation

FestiveKyle:
Fixes "error" on sign up (removed affiliations field from mutation)


tparrott-cse (approved):
LGTM!

PR #1475, opened 2021-01-18
Fix incorrect password prompt not appearing

ThomasNickerson:
Fix error handling for sign in mutation to allow incorrect password prompt to appear.


tparrott-cse (approved):
LGTM!

PR #1481, opened 2021-01-19
Fe add message when no scan data for org

ThomasNickerson:
- Adds a message and logo that are displayed when there is no scan data to generate a doughnut chart from. 
- Updates associated test to allow for localized text.
- Updates associated test to use data that is more like real data (added missing total attribute).


lcampbell2 (dismissed):
Looks great! One small thing that needs to be updated is the translation for DMARC pass/fail. Make sure the text in en.po and fr.po matches exactly with what's in the code

ThomasNickerson:
> Looks great! One small thing that needs to be updated is the translation for DMARC pass/fail. Make sure the text in en.po and fr.po matches exactly with what's in the code

Thanks! The DMARC pass/fail strings should be translating correctly as of the most recent commit

ThomasNickerson:
Rebased onto master for lingui update at Kyle's suggestion

tparrott-cse (approved):

PR #1548, opened 2021-02-01
Client implement first tests

ThomasNickerson:
This PR implements test coverage for all formatting functions, as well as placeholder tests for some connectivity related functions. Query strings were also moved out of the main client module.


nsdeschenes (changes requested):
I'm wondering if you're able to make the change from `src/` to `tracker_client/`, the `src/` is JS convention whereas with python its based off of the project/module name. More info on the conventions here: [sample repo](https://github.com/navdeep-G/samplemod) and [docs](https://docs.python-guide.org/writing/structure/)

ThomasNickerson:
> I'm wondering if you're able to make the change from `src/` to `tracker_client/`, the `src/` is JS convention whereas with python its based off of the project/module name. More info on the conventions here: [sample repo](https://github.com/navdeep-G/samplemod) and [docs](https://docs.python-guide.org/writing/structure/)

For sure, I'll take a look and make the change. What I had planned on originally was the src/pkg_name layout described in [pytest docs](https://docs.pytest.org/en/stable/goodpractices.html#test-discovery), which refer to this [blog post](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure), but I was having trouble with imports and decided to simplify temporarily. Do you think it's worth considering that layout instead?

tparrott-cse:
I think it's probably best to stick with the Python conventions for the client.

nsdeschenes (approved):

tparrott-cse (approved):

PR #1567, opened 2021-02-03
Start centralizing logging

sleepycat:
This is a first step to centralized logging. GCP uses structured logs,
essentially json objects, which let you capture every detail and filter for the
ones that matter.
GCP has it's [own structure](https://bit.ly/3oGwtuI) as will other platforms
and creating a [central point to manage this stuff](https://bit.ly/2O0oBaK), as
well as filter out certain details before they get sent to the provider, will
let us make the most of the providers infrastructure.

As a bonus, this approach helps us with our tests too: passing a logger object
in makes it easy to mock for testing, where console.log statements force us
into heavy handed replacement of the global console object.

### DIFF
--- a/api-js/src/create-context.js
+++ b/api-js/src/create-context.js
@@ -295,20 +295,21 @@ export const createContext = ({ context, req: request, res: response }) => {
       ),
       orgLoaderByKey: orgLoaderByKey(query, request.language, userKey, i18n),
       orgLoaderBySlug: orgLoaderBySlug(query, request.language, userKey, i18n),
-      orgLoaderConnectionArgsByDomainId: orgLoaderConnectionArgsByDomainId(
+      orgLoaderConnectionArgsByDomainId: orgLoaderConnectionArgsByDomainId({
         query,
-        request.language,
+        language: request.language,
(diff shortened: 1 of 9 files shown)


api-js/src/create-context.js
```diff
@@ -295,20 +295,21 @@ export const createContext = ({ context, req: request, res: response }) => {
       ),
       orgLoaderByKey: orgLoaderByKey(query, request.language, userKey, i18n),
       orgLoaderBySlug: orgLoaderBySlug(query, request.language, userKey, i18n),
```
nsdeschenes, line 297: Are you going to update all loaders, or just the connection ones?

api-js/src/create-context.js
```diff
@@ -295,20 +295,21 @@ export const createContext = ({ context, req: request, res: response }) => {
       ),
       orgLoaderByKey: orgLoaderByKey(query, request.language, userKey, i18n),
```
nsdeschenes, line 296: Are you going to update all loaders, or just the connection ones?

sleepycat:
Closed as outdated.

PR #1648, opened 2021-02-15
Client improve test coverage

ThomasNickerson:
This PR implements a minimal set of unit tests for 100% code coverage. Also ~~added URL as a parameter with a default set to the alpha tracker endpoint in cases where that value was hardcoded, and~~ (scratch that, was in final commit of previous PR)  improved consistency of query variable names in the case of get_domains_by_name.


nsdeschenes (approved):

tparrott-cse (approved):
Great work!

PR #1732, opened 2021-02-24
Client fix guidance

ThomasNickerson:
This PR updates the client for the deprecation of the guidanceTags field.

### DIFF
--- a/clients/python/tracker_client/formatting.py
+++ b/clients/python/tracker_client/formatting.py
@@ -118,7 +118,7 @@ def format_name_summary(result):
     }
     return result
 
-
+# TODO: refactor this to reduce repetition
 def format_all_results(result):
     """Formats the result dict in get_all_results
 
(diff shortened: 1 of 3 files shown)


clients/python/tracker_client/formatting.py
```diff
@@ -140,29 +141,73 @@ def format_all_results(result):
     # Extract the contents of the list of edges for guidance tags
     for x in result["findDomainByDomain"]["web"].keys():
```
tparrott-cse, line 142: Can you make "x" more descriptive?

ThomasNickerson: For sure, its value is "https", "ssl", "dmarc" etc, so something like scan_category or result_category?

tparrott-cse (approved):

clients/python/tracker_client/formatting.py
tparrott-cse: yup, your suggestion works fine

nsdeschenes (approved):

PR #1754, opened 2021-03-01
Client packaging

ThomasNickerson:
This PR adds license information and a setup.py file to allow the client to be uploaded to, and installed from, PyPI. Also adds packaging tools (setuptools, wheel, twine) to the Pipfile dev requirements, and fixes version requirements for gql in the Pipfile.


clients/python/setup.py
```diff
@@ -0,0 +1,33 @@
...
+
+setuptools.setup(
+    name="tracker_client",
+    version="1.0.0-a",
+    author="Thomas Nickerson",
+    author_email="contact@cyber.gc.ca",
+    description="A Python client for the GoC Tracker API",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/canada-ca/tracker",
```
nsdeschenes, line 20: Because we have a mono repo it may be better to directly link to the clients directory: `https://github.com/canada-ca/tracker/tree/master/clients/python`

nsdeschenes (changes requested):
Looking good, minus this one little nitpick

clients/python/setup.py
ThomasNickerson: Makes sense, I’ll make the change. Thanks!

nsdeschenes (approved):

tparrott-cse (approved):

PR #1811, opened 2021-03-08
Client deal with pagination

ThomasNickerson:
This PR updates `Client.get_domains`, `Client.get_organizations` and `Organization.get_domains` to use pagination to get as many items as exist, rather than just the first 100.

### DIFF
--- a/clients/python/tracker_client/queries.py
+++ b/clients/python/tracker_client/queries.py
@@ -130,10 +130,6 @@ GET_DOMAIN_OWNERS = gql(
     query FindDomainByDomain($domain: DomainScalar!) {
         findDomainByDomain(domain: $domain) {
             organizations(first: 100) {
-                pageInfo{
-                    hasNextPage
-                    endCursor
-                }
                 edges {
                     node {
                     acronym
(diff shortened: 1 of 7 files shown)


tparrott-cse (approved):

tparrott-cse (approved):

clients/python/tracker_client/queries.py
```diff
@@ -94,8 +98,12 @@
 # Get scalar fields of all your orgs
 GET_ALL_ORGS = gql(
     """
-    query GetAllOrgs {
-        findMyOrganizations(first: 100) {
+    query GetAllOrgs($after: String) {
```
nsdeschenes, line 101: wondering if this should be `getAllOrgs($after: String)`, to match other operation names in this file.

nsdeschenes (changes requested):

clients/python/tracker_client/queries.py
ThomasNickerson: [In the queries for the frontend](https://github.com/canada-ca/tracker/blob/016ced8d44053fae2b15cda6fdd2f3e9564fd434/frontend/src/graphql/queries.js) query names all begin with caps, do you think it would be better to change everything to match that?

PR #1812, opened 2021-03-08
Client handle login error

ThomasNickerson:
This PR updates `get_auth_token` to not swallow errors passed back in the SignInUnion and instead raise an exception with error details. Also changed `create_transport` to `_create_transport` as there is no reason to access it from outside core.py.

### DIFF
--- a/clients/python/tracker_client/queries.py
+++ b/clients/python/tracker_client/queries.py
@@ -29,6 +29,10 @@ SIGNIN_MUTATION = gql(
                 ... on AuthResult {
                     authToken
                 }
+                ... on SignInError{
+                    code
+                    description
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


clients/python/tracker_client/queries.py
```diff
@@ -29,6 +29,10 @@
                 ... on AuthResult {
                     authToken
                 }
+                ... on SignInError{
```
nsdeschenes, line 32: This may be better off in a separate PR, however we may want to add support in for users who have tfa enabled.
```graphql
result {
  ... on TFASignInResult {
    sendMethod
    authenticateToken
  }
  ... on AuthResult {
    authToken
  }
  ... on SignInError {
    code
    description
  }
}
```

nsdeschenes (changes requested):

clients/python/tracker_client/queries.py
ThomasNickerson: I'll add it to the project board, I think that should definitely be added but in another PR.

tparrott-cse (approved):

nsdeschenes (approved):

PR #1816, opened 2021-03-08
Do not fetch schema with Python GraphQL

Ethanljf:
This change addresses an issue where an error is encountered if `fetch_schema_from_transport=True`

Not really sure why... but that's just the way it goes sometimes :/


sleepycat (approved):
LGTM!

PR #1834, opened 2021-03-09
Small Fix For updateUserProfile Mutation

nsdeschenes:
Add in a check to ensure username is not already in use.

### DIFF
--- a/api-js/src/user/mutations/update-user-profile.js
+++ b/api-js/src/user/mutations/update-user-profile.js
@@ -48,7 +48,7 @@ export const updateUserProfile = new mutationWithClientMutationId({
       i18n,
       query,
       userKey,
-      loaders: { userLoaderByKey },
+      loaders: { userLoaderByKey, userLoaderByUserName },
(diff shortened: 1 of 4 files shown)


api-js/src/user/mutations/update-user-profile.js
```diff
@@ -77,6 +77,17 @@ export const updateUserProfile = new mutationWithClientMutationId({
...
     }
 
+    // Check to see if user name is already in use
+    if (userName !== '') {
+      const checkUser = await userLoaderByUserName.load(userName)
+      if (typeof checkUser !== 'undefined') {
+        console.warn(
+          `User: ${userKey} attempted to update their username, but the username is already in use.`,
+        )
+        throw new Error(i18n._(t`Unable to update profile. Please try again.`))
```
tparrott-cse, line 87: We might need to be more specific than this -- a user might think this is a system error and not something they need to address.

Ethanljf (approved):

tparrott-cse (approved):

PR #1872, opened 2021-03-15
Fe fix adding phone number

### LINKED ISSUE #1686: Phone number isn't displayed after adding it
Adding a phone hits the API, but doesn't update the UI.
Screenshot from 2021-02-19 18-07-58

Reloading the app, the API reports the phone number as `null`.

Screenshot from 2021-02-19 18-13-49

ThomasNickerson:
This PR fixes the ability to add/edit and verify user phone numbers. This includes:
- adding mutation strings for `setPhoneNumber` and `verifyPhoneNumber`
- updating `EditableUserPhoneNumber` to use `setPhoneNumber` rather than `updateUserProfile`
- adding a second modal dialog and associated logic to verify phone number after adding/changing it
- small addition to `AuthenticateField` to allow its reuse in the above
- updating locale files

Closes #1686.


lcampbell2 (approved):
:fire: :fire: :fire:

tparrott-cse (approved):

FestiveKyle (approved):

PR #1878, opened 2021-03-16
Configure Dependabot V2

ThomasNickerson:
Adds a `dependabot.yml` config file to tame dependabot slightly and switch to the newer version.

Configuration is based on the automatically generated file from PR #897 with the open PR limit changed to 10 (from 99) and updated to the repo's current structure. It may be worth tweaking options more, so please suggest changes or make them yourself if you think of anything. You can find information about possible options [here](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,184 @@
+version: 2
+updates:
+- package-ecosystem: docker
+  directory: "/api-js"
+  schedule:
+    interval: daily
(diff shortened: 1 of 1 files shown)


sleepycat (approved):
:star_struck: Looks great! Finally taming dependabot!

tparrott-cse (approved):

PR #1975, opened 2021-03-26
SSL results: unknown -> False for evaluation purposes

Ethanljf:
These changes update the result format for SSL to use "False" rather than unknown for evaluation purposes.


nsdeschenes (approved):
:fire:

PR #2013, opened 2021-03-31
Update sslyze and minor refactor of SSL scanner

ThomasNickerson:
This PR:
- Updates sslyze to 4.1.0 from 3.1.0 for bug fixes and stability improvements.
- Removes preferred cipher suite detection as results were not used anywhere. This capability was removed from sslyze for unreliability and the use of it was forcing us to stay on sslyze 3.1.0.
- Includes minor refactoring:
    - Ran Black.
    - The function `get_server_info` was conditionally returning `None` based on an exception being raised, and its return type was used for control-flow. Refactored so that the exception is caught by the caller instead and the function always returns the same type OR raises an exception.

### DIFF
--- a/services/scanners/ssl/requirements.txt
+++ b/services/scanners/ssl/requirements.txt
@@ -4,7 +4,7 @@ uvicorn
 emoji
 pytest
 requests>=2.18.4
-sslyze==3.1.0
+sslyze==4.1.0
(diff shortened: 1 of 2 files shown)


Ethanljf (approved):
🤩

PR #2083, opened 2021-04-09
Align dmarc summary order fields

FestiveKyle:
As described


nsdeschenes (approved):
Careful here, you're gonna put me out of a job at this rate :joy_cat:

PR #2103, opened 2021-04-13
Update Org Loader and Query to Support Admin Filter

nsdeschenes:
This PR adds a new `isAdmin` filter to the `findMyOrganizations` query filtering the orgs for ones that the given user is an admin or higher to.


Ethanljf (approved):

sleepycat (approved):
LGTM! :shipit:

PR #2104, opened 2021-04-13
Fix scanner image updates

ThomasNickerson:
Flux was set to check the wrong GCR directory for updated scanner images, this is now fixed.


sleepycat (approved):
:star_struck:

PR #2121, opened 2021-04-16
Fe add search and ordering to dmarc summaries

### LINKED ISSUE #2128: Date changer on DMARC summary page doesn't work
Changing the date modifying does not cause a network request for the new date.

Fix:
Modify cache to make date keyfields

FestiveKyle:
As described. Also:
- Changes loading spinner design
- Removes "no sort" option for dmarc summary table (2 options for each column instead of 3)
- Adds useDebounce, helpful for reducing the number of wasteful fetches
- Closes #2128


lcampbell2 (approved):

frontend/src/helpers/toConstantCase.js
```diff
@@ -0,0 +1,5 @@
+import { upperCase } from 'lodash'
+
+export const toConstantCase = (str) => {
+  return upperCase(str).replace(/ /g, '_')
```
sleepycat, line 4: Let's not create a dependency on Lodash for this. If we change this to `str.toUpperCase().replace(/ /g, '_')`, it'll give us the same behaviour without dragging in new a dependency.

sleepycat (changes requested):
One minor change! The rest looks good!

frontend/src/helpers/toConstantCase.js
FestiveKyle: Done 👍 

```
export const toConstantCase = (str) => {
  return str.toUpperCase().replace(/ /g, '_')
}

```

PR #2123, opened 2021-04-16
Fe always show phone number help

### LINKED ISSUE #2122: Phone number format guidance only visible in validation error message
Editing your phone number opens this dialog box. It's unclear what format I should use, (`555-555-5555`? `5555555555`? etc) and no guidance is shown.
Screenshot from 2021-04-16 10-21-36

It turns out there is useful guidance, but it's only shown as a validation error:
Screenshot from 2021-04-16 10-21-16

We should consider adding placeholder text in the format we want, and/or taking the message we display when validation fails and making it visible all the time.

FestiveKyle:
As described.

Closes #2122 

Screenshot (27)

### DIFF
--- a/frontend/src/PhoneNumberField.js
+++ b/frontend/src/PhoneNumberField.js
@@ -35,14 +35,6 @@ const PhoneNumberField = WithPseudoBox(function PhoneNumberField({
       ''
     )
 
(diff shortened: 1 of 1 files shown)


sleepycat (approved):
:star_struck:

PR #2180, opened 2021-05-06
Document autoscan and queues

ThomasNickerson:
This PR adds docstrings to the Python source for Autoscan, the scan queue and result queue, and READMEs to all of their root directories.


Ethanljf (approved):
This is absolutely amazing! Great work!

This was a much-needed addition.

PR #2194, opened 2021-05-10
Rework mocking for API

FestiveKyle:
Introduce graphql-tools mocking as a replacement for graphql-faker. 

Can be run with `npm run mocker` or a containerized version using docker-compose. Both versions allow server auto-restarts when modifying code in the mocking/  directory


sleepycat (approved):
LGTM! :tada:

lcampbell2 (approved):
:fire: :fire: :fire:

nsdeschenes (approved):

PR #2241, opened 2021-05-18
Add 'Accept-Language' header to requests on frontend

FestiveKyle:
- Sets the 'Accept-Language' header according to the current locale
- Resets the cache on language switch
- Sets the frontend language to the browser's language by default


sleepycat (approved):
LGTM :+1:

PR #2249, opened 2021-05-19
Refetch queries when removal/modification of domain

### LINKED ISSUE #2244: UI not updating when domains are removed
Removing a domain succeeds on the server but the UI is not updated.
Here I removed the domain `arangodb.com` and while the notification indicates success (it does actually remove it!) the UI is still displaying the domain.
image

FestiveKyle:
As described

Closes #2244


sleepycat (approved):
:star_struck:

PR #2251, opened 2021-05-19
Add Super Admin Org Filter

nsdeschenes:
This PR adds in `includeSuperAdminOrg` a boolean filter argument to the `findMyOrganizations` query, and `organizations` field on the `Domain` GQL object.


sleepycat (approved):
:+1:

PR #2256, opened 2021-05-20
Update phone number on frontend after number is set

### LINKED ISSUE #2125: Phone number isn't displayed after update
Updating your phone number works, but does not display anything to the user. 

phone-number-update

After logging out and logging back in, the number is correctly displayed.

phone-number-displayed

FestiveKyle:
- Add user to return object for setPhoneNumber
- Update set-phone-number.test.js to test for updated phone number on user
- Update phone number for user on frontend after user sets new number
- Update mocking

Closes #2125


lcampbell2 (approved):
:fire:

api-js/src/user/objects/user-personal.js
```diff
@@ -10,6 +10,19 @@ import { nodeInterface } from '../../node'
 
 const { CIPHER_KEY } = process.env
 
+export const decryptPhoneNumber = ({iv, tag, phoneNumber: encrypted}) => {
```
nsdeschenes, line 13: Can you put this back where it was, it's the only location that its being used, so I'd prefer to keep it inside the resolver itself.

nsdeschenes (changes requested):

api-js/src/user/objects/user-personal.js
FestiveKyle: I pulled this function out so we can use it to test for changed phone numbers like here: https://github.com/canada-ca/tracker/blob/5abafc0aac45cf9e39332fd2f6d62efc9506e996/api-js/src/user/mutations/__tests__/set-phone-number.test.js#L165

nsdeschenes: okay, then remove it from the file, and create it, its own folder in the validators directory and add it to context, if you need help lmk.

nsdeschenes (approved):

PR #2301, opened 2021-05-28
Fix formatting of domain cards

FestiveKyle:
Fix DomainCard from overflowing on smaller screens. Also switches breakpoints for some properties (such as the scan result circle's flexDirection) and aligns the scan result circles.

Before:
image

After:
image

Before:
image

After:
image

Before:
image
image

After:
image


lcampbell2 (approved):
:fire: :fire: :fire:

PR #2322, opened 2021-05-31
Allow for organization summaries to be null on org summary page

FestiveKyle:
Adds text in place of summary cards when summary data does not exist.

image


sleepycat (approved):
:+1:

PR #2328, opened 2021-06-01
FE Change Network Policy for Queries with Search

FestiveKyle:
Changes the fetchPolicy for queries with search to 'cache-and-network', and sets nextFetchPolicy to 'cache-first'.

This behaviour causes the first instance of a query (with its specific arguments) to send a network request. Setting the nextFetchPolicy to 'cache-first' will ensure that subsequent requests for the same query can be pulled from the cache. This cycle is destroyed on component unmount or if the query changes. In practice, this means that changing search/order or switching off the page and coming back will ensure the user gets fresh data.


lcampbell2 (approved):

sleepycat (approved):
:tada:

PR #2352, opened 2021-06-03
Fixes language button jumping on hover

AdamPrins:
adds a transparent border so that the size doesn't change on hover


AdamPrins:

PR #2354, opened 2021-06-03
Reverse order of create account buttons

AdamPrins:
Create account was on the left, and the back button was on the right. This original ordering is unintuitive.


AdamPrins:

PR #2355, opened 2021-06-03
add report button to table

AdamPrins:
The text was a hyperlink, but not readily visible in anyway.
I only noticed it was a link by accident. This button is a much clearer indicator to navigating to the DMARC Report


AdamPrins:

PR #2362, opened 2021-06-04
add report button to table

AdamPrins:
The text was a hyperlink, but not readily viable in anyway.
I only noticed it was a link by accident. This button is a much clearer indicator to navigating to the DMAR Report page

### DIFF
--- a/frontend/src/DmarcByDomainPage.js
+++ b/frontend/src/DmarcByDomainPage.js
@@ -24,7 +24,6 @@ import { usePaginatedCollection } from './usePaginatedCollection'
 import { RelayPaginationControls } from './RelayPaginationControls'
 import { toConstantCase } from './helpers/toConstantCase'
 import { useDebounce } from './useDebounce'
-import { TrackerButton } from './TrackerButton'
 import { Link as RouteLink } from 'react-router-dom'
 
 export default function DmarcByDomainPage() {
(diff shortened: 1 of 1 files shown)


frontend/src/DmarcByDomainPage.js
```diff
@@ -115,6 +117,24 @@ export default function DmarcByDomainPage() {
...
+              as={RouteLink}
+              to={`/domains/${value}/dmarc-report/LAST30DAYS/${new Date().getFullYear()}`}
+            >
+              <Text whiteSpace="noWrap">
+                <Trans>DMARC Report</Trans>
+              </Text>
+            </TrackerButton>
+          </Stack>
+        )
+      },
```
FestiveKyle, line 137: I think the following code is a good compromise of being recognizable as a link but still remaining subtle. It should also fix the eslint issue.

```
      import { Link as RouteLink } from 'react-router-dom'
      
      // eslint-disable-next-line react/prop-types
      Cell: function CellValueWithLink({ value }) {
        return (
          <Link
            as={RouteLink}
            to={`domains/${value}/dmarc-report/LAST30DAYS/${new Date().getFullYear()}`}
            isExternal={false}
          >
            {`${value} `} <Icon name="link" />
          </Link>
        )
      },
```

image

FestiveKyle:
Looks like it just needs Link imported from '@chakra-ui/core' and it should be good

FestiveKyle (approved):

PR #2373, opened 2021-06-04
Sign in and Create account buttons on home page

AdamPrins:
Sign in is now a TrackerButton
Create Account TrackerButton is added beside the sign in button

Uses new TrackerButton variant from #2372


AdamPrins:
Being replaced by #2456

PR #2375, opened 2021-06-04
Fix Sign-out button ending up on its own line for small widths

AdamPrins:
This change makes it so that the Sign out button no longer ends up on its own line for medium sizes.


sleepycat (approved):
:+1:

PR #2378, opened 2021-06-04
Ux reference dmarc summaries by name

AdamPrins:
paths to /dmarc-summaries used DMARC Report, and the summary table was labeled DMARC Messages.
These are all now referenced as DMARC Summaries


AdamPrins:
Renames the floating menu from  #2376

AdamPrins:
Rebased to resolve conflict.

PR #2379, opened 2021-06-04
change signup error from referencing Username to Email

AdamPrins:
There is no field the user can see called username,
So they were not clearly told it was the email that was causing the issue.


sleepycat (approved):
Nice!

PR #2392, opened 2021-06-07
Reference /dmarc-summaries by name

AdamPrins:
paths to /dmarc-summaries used DMARC Report, and the summary table was labeled DMARC Messages.
These are all now referenced as DMARC Summaries

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -77,7 +77,7 @@ export default function App() {
 
           {isLoggedIn() && (
             <Link to="/dmarc-summaries">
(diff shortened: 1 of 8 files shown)


FestiveKyle (changes requested):
Just needs the translations added to the \*.po files! (the locales/\*.js files are automatically compiled from the \*.po files)

AdamPrins:
done, didn't realize that I should be editing the .po and not the .js locales files.

FestiveKyle (approved):

PR #2394, opened 2021-06-07
Allow super admin to invite super admins

### LINKED ISSUE #2393: Super Admin Can't Create More Super Admins
**Describe the bug**
When inviting users to the Super Admin org, they can't invite users with the super admin role.

**To Reproduce**
1. Go to 'Admin Profile'
2. Select 'SA'
3. Click 'Users'
4. Select 'Role List'

**Expected behavior**
If the user is a super admin, and they are viewing the super admin org, they should be able to invite users at the super admin level.

FestiveKyle:
Allows a super admin to request an invite **only** at the super admin level (only for the super-admin org). Also only allows changing another user's role to SUPER_ADMIN if the change is for the super-admin org.

Closes #2393 

image

### DIFF
--- a/frontend/src/UserList.js
+++ b/frontend/src/UserList.js
@@ -359,8 +359,12 @@ export default function UserList({ permission, orgSlug, usersPerPage, orgId }) {
               flexShrink={0}
               {...userForm.getFieldProps('roleSelect')}
             >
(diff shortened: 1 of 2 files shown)


sleepycat (approved):
:fire: Awesome!

nsdeschenes (approved):

PR #2395, opened 2021-06-08
Fix Super Admin Org Filter

nsdeschenes:
This PR fixes changes with the new slug for the super admin organization so that it can be filtered in the admin profile org list.


FestiveKyle (approved):
👍

PR #2407, opened 2021-06-09
Fix user role value for super-admin org

FestiveKyle:
Fixes the initial value set in the invitation area on the admin page. The super-admin org does not have a 'USER' option, but the initial value was still set to 'USER'.


nsdeschenes (approved):
:shipit:

PR #2409, opened 2021-06-09
fixes #2399

### LINKED ISSUE #2399: Web/Email guidance are different widths
Web and email guidance have different widths, and switching between the is jarring.

width-change

AdamPrins:
Makes it so that the Web and Email guidance are the same width.


FestiveKyle (commented):
I think isFitted is a special prop for the Tabs component. We should be good with maxW='100%' or width='100%'

FestiveKyle (approved):

PR #2415, opened 2021-06-10
FE add curves to guidance page

lcampbell2:
added curves to SSL category on Web Scan card


FestiveKyle (changes requested):
Just needs translations 👍

FestiveKyle (approved):

PR #2425, opened 2021-06-11
Add Help Page

AdamPrins:
Adds a help page with some documentation, and links for further assistance.


sleepycat:
Screenshot from 2021-06-11 16-37-50

sleepycat:
The help page links point to the following documents:
https://www.gcpedia.gc.ca/gcwiki/images/f/f9/Web_Sites_and_Services_Management_Configuration_Requirements_-_20210517.pdf
https://www.gcpedia.gc.ca/gcwiki/images/d/d0/Email_Management_Services_Configuration_Requirements_-_20210517.pdf

Tracker web and email guidance pages exist to help with those things, linking to guidance and giving specific advice on fixes.
Pointing people to these documents suggests that Tracker has failed at it's job. If there is useful content in those documents we should be folding it into those pages and showing it with the rest of our guidance stuff.

The links themselves are also problematic: GCpedia links don't work for non-GC users, and PDFs are [notoriously awkward on the web](https://gds.blog.gov.uk/2018/07/16/why-gov-uk-content-should-be-published-in-html-and-not-pdf/).
Let's put out heads together and figure if generic help still makes sense or how to give people the help they need in context instead.

PR #2426, opened 2021-06-11
add display names to UserCard

AdamPrins:
Adds each users displayName to the User Card
Screen Shot 2021-06-11 at 12 05 25 PM

They display Hello World on the local build, but I'm not sure how to check the data structure that we pull from, or edit it so that they have proper display names for testing.

### DIFF
--- a/frontend/src/OrganizationAffiliations.js
+++ b/frontend/src/OrganizationAffiliations.js
@@ -59,6 +59,7 @@ export function OrganizationAffiliations({ usersPerPage = 10, orgSlug }) {
             >
               <UserCard
                 userName={user.userName}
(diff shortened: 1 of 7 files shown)


AdamPrins:
also, for OrganizationAffiliations.test.js I added another test, but I'm not sure what our policy on tests is.

Should displayNames get a test?
And if so, should it be in its own await like I have it, or should it be inside the original await with the other expect?

sleepycat (approved):
:+1: Good idea!

sleepycat:
For testing, we're interested in behaviour, but we assume that React's behaviour is covered by it's own tests.

PR #2439, opened 2021-06-14
Update EditableUserPhoneNumber.js

AdamPrins:
Previous regex expression `^\+[1-9]\d{10,15}$` matches a + sign, a digit between 1 and 9, and then 10 to 15 more digits.

This made it unable to accept phone numbers of length 10, and able to accept phone numbers of length 16. Acceptable lengths are supposed to be from 10-15 inclusive.


sleepycat (approved):
:+1:

PR #2442, opened 2021-06-14
Right-align domains in domain cards when not scanned

FestiveKyle:
Domain names were not right-aligned when not scanned. This fix allows the domain name area to FlexGrow to fill the rest of its parent's card width.

Before:
image

After:
image


lcampbell2 (approved):
:eagle: :eye:

PR #2443, opened 2021-06-14
Add check for certificate revocation to HTTPS scanner

ThomasNickerson:
This PR adds checking for revoked certificates to the HTTPS scanner using Mozilla's [CRLite](https://blog.mozilla.org/security/2020/01/09/crlite-part-2-end-to-end-design/).

This consists of a Python function that checks a cert against CRLite, modifications to the existing HTTPS scanner to get and process this result, and two new guidance tags for cert revocation.


Ethanljf (approved):
This is awesome! Great work🤩

PR #2450, opened 2021-06-15
Change phone field

AdamPrins:
Uses a react module for auto phone formatting. This makes it easier for users to input their phone number


AdamPrins:
Error message for empty field works, message for invalid input (too short) does not.

No substantial changes were made to `phoneValidationSchema` in `EditableUserPhoneNumber.js`, so cause is unknown.

Screen Shot 2021-06-15 at 10 32 41 AM

Screen Shot 2021-06-15 at 10 32 48 AM

frontend/src/PhoneNumberField.js
```diff
@@ -1,28 +1,24 @@
...
-  Icon,
-  Input,
-  InputGroup,
-  InputLeftElement,
 } from '@chakra-ui/core'
-import { useField } from 'formik'
+import { useField, useFormikContext } from 'formik'
 import WithPseudoBox from './withPseudoBox'
 import { fieldRequirements } from './fieldRequirements'
+import PhoneInput from 'react-phone-input-2'
```
sleepycat, line 14: Can you take a stab at lazy loading this library? I remember it's pretty big (a massive improvement on libphonenumber-js but still big). Try this:

```js 
const PhoneInput = lazy(() => import('react-phone-input-2))
```
It's a default export, so I think [React.lazy](https://reactjs.org/docs/code-splitting.html#reactlazy) would work here. The result would be the JS bundle is split up more and chunks loaded as needed.
If it works, let's do that.

sleepycat (changes requested):
This looks good. Just try switching this to lazy load and if that works we'll get this in.

PR #2466, opened 2021-06-18
Fe add loading state to buttons

### LINKED ISSUE #2459: No loading indication when adding a new org
When adding a new org, there is no indication that the request was sent and the user should await a response. 

To fix:
TrackerButton accepts an isLoading state.

FestiveKyle:
Adds loading state to the "Add Domain" and "Invite User" buttons

Closes #2459


sleepycat (approved):
:star_struck:

PR #2468, opened 2021-06-18
remove create account button

AdamPrins:
Create account button no longer needed on sign in page now that it is in the top bar.

Screen Shot 2021-06-18 at 10 19 36 AM


sleepycat (approved):
Good catch!

Issue #2475, opened 2021-06-21
Title Bar Sign In/Create Account Buttons on Mobile Version


nsdeschenes:
**Describe the bug**
Looking at the mobile site, the `Sign In` and `Create Account` buttons seem sort of small, and we already have a sign in button in the floating menu. I'm suggesting that we remove the buttons in the title bar, and add in a `Create Account` button in the floating menu, making it easier to navigate on the mobile site.

**Screenshots**
<img src="https://user-images.githubusercontent.com/7891920/122741074-fa1c9980-d25a-11eb-8d7e-858bb8ccec3f.jpg" width="300">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<img src="https://user-images.githubusercontent.com/7891920/122741082-fb4dc680-d25a-11eb-83bf-9f7a8b0df4a7.jpg" width="300">

AdamPrins:
Create account added to floating menu.
Buttons in top bar hide for small width

PR #2480, opened 2021-06-21
add create account button to floating menu

AdamPrins:
implements the recommendations in #2475

Hides sign in buttons on mobile width, since they are located in the floating menu


sleepycat (approved):
Nice! :shipit:

PR #2486, opened 2021-06-21
Fix KeyError when domain has no dmarc record

ThomasNickerson:
This PR fixes an error caused by the DNS scanner attempting to examine the DMARC record of domains that lack one.


Ethanljf (approved):
LG(reat!)TM

PR #2498, opened 2021-06-23
Implement New Owner Field on Affiliations

nsdeschenes:
This new owner field on affiliation connections, allows us to see who the owner of a given sandbox organization is. This field is set to true to the user who created the org, and eventually will be able to be transferred to another user. 

This field is also added in when an existing user is invited to an org, or when they sign up with a token, the field is set to false.

When a super admin verifies an org, all owner fields are set to false.


Ethanljf (approved):

sleepycat (approved):
:+1: :ship:

PR #2499, opened 2021-06-23
Ux user card overflow

### LINKED ISSUE #2460: UserCard: Display names cause permissions to overflow on smaller windows (mobile)
On mobile the display names are causing permission roles on the right to overflow and the emails on the left to line break

Screenshot from 2021-06-17 14-14-39

AdamPrins:
fixes #2460

User Card now uses rows for mobile width

Stack doesn't support responsive styling for direction in 0.8
So There's some responsive display props.


AdamPrins:
Replaced by  #2516

PR #2501, opened 2021-06-23
Fix Sorting Issues on Connection Loaders

nsdeschenes:
Currently there are issues with sorting the data coming from connection loaders because keys weren't being sent to number.


FestiveKyle (approved):
🔨

Ethanljf (approved):

PR #2504, opened 2021-06-23
Fe move user state to reactive var

FestiveKyle:
Switches from the useUserState state-based hook to useUserVar reactive-variable-based hook. 

- The function calls dealing with the user state are identical to the old implementation
- UsesApollo 3's reactive variables for managing state (separate from cache)
- Allows access of the user state property outside of the React application which is needed for subscriptions
- Adding the 'authorization' header on a case-by-case basis (ex. during useQuery) is no longer required


lcampbell2 (approved):

sleepycat (approved):
:star_struck:

PR #2505, opened 2021-06-23
Remove id query from ORG_DETAILS_PAGE

AdamPrins:
Exposes IDs to accounts with only user permissions in an org
significantly increases query response size
Results were not being used


AdamPrins:
Replaced by #2518

PR #2508, opened 2021-06-24
Update isUserAdmin to Support Individual Org Checks

nsdeschenes:
This PR adds a new arg onto the `isUserAdmin` query to support checking if the user requesting user is an admin for a specific org.

```graphql
query {
  isUserAdmin (orgId: "")
}

query {
  isUserAdmin (orgId: "")
}
```


AdamPrins (approved):
I don't know the most about the api, but it looks good

sleepycat (approved):
LGTM! :shipit:

PR #2512, opened 2021-06-24
Update models.py to add required attribute

Ethanljf:
Resolves issue related to accessing non-existent endpoint attribute.

To replicate, scan `aanc-inac.gc.ca` for HTTPS configuration

`'Endpoint' object has no attribute 'https_cert_revoked'`


ThomasNickerson (approved):
Nice catch!

PR #2516, opened 2021-06-25
Ux user card overflow rebase

### LINKED ISSUE #2460: UserCard: Display names cause permissions to overflow on smaller windows (mobile)

AdamPrins:
fixes #2460

User Card now uses rows for mobile width

Stack doesn't support responsive styling for direction in 0.8
So There's some responsive display props.

### DIFF
--- a/frontend/src/UserCard.js
+++ b/frontend/src/UserCard.js
@@ -5,14 +5,32 @@ import { string } from 'prop-types'
 export function UserCard({ userName, displayName, role }) {
   return (
     <PseudoBox width="100%" p="8">
-      <Stack isInline align="center" mb={['1', '0']}>
-        <Box width="50%">
+      <Box
+        align="center"
+        mb={['1', '0']}
+        width="100%"
+      >
+        <Box
+          width={{sm:"100%", md:"50%"}}
+          display={{sm:"block", md:"inline-block"}}
+          align={{sm:"left", md:"center"}}
+        >
           <Text fontSize="md" wordBreak="break-all">
             {userName}
           </Text>
         </Box>
-        <Box width="40%">{displayName}</Box>
-        <Box width="10%">
+        <Box
+          width={{sm:"100%", md:"40%"}}
+          display={{sm:"block", md:"inline-block"}}
+          align={{sm:"left", md:"center"}}
+        >
+          {displayName}
+        </Box>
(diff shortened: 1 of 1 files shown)


FestiveKyle (changes requested):
This still lets components over-flow their parents. 
image

Using a grid here would let everything still be aligned in columns and have the ability to wrap in case of long email addresses on smaller screens. Something like this in OrganizationAffiliation would work well:
```
  return (
    <ErrorBoundary FallbackComponent={ErrorFallbackMessage}>
      <Grid
        grid-template-dolumns={{ base: 'auto', md: 'auto auto auto' }}
        columnGap="1.5rem"
        rowGap={{ md: '0.5rem' }}
      >
        {nodes.map(({ permission, user }) => {
          return (
            <React.Fragment key={user.displayName}>
              <Text fontSize="md" wordBreak="break-all" ml={{ md: '1rem' }}>
                {user.userName}
              </Text>
              <Text>{user.displayName}</Text>
              {permission && (
                <Badge
                  color="primary"
                  bg="transparent"
                  borderColor="primary"
                  borderWidth="1px"
                  alignSelf={{ md: 'center' }}
                  mr={{ md: '1rem' }}
                  justifySelf={{ base: 'start', md: 'end' }}
                >
                  {permission}
                </Badge>
              )}
              <Divider
                borderColor="gray.900"
                gridColumn={{ base: 'span 1', md: 'span 3' }}
              />
            </React.Fragment>
          )
        })}
      </Grid>

      <RelayPaginationControls
        onlyPagination={true}
        hasNextPage={hasNextPage}
        hasPreviousPage={hasPreviousPage}
        next={next}
        previous={previous}
        isLoadingMore={isLoadingMore}
      />
    </ErrorBoundary>
  )

```

FestiveKyle:
Using grid with the individual elements (userName, displayName, permission) as direct children allows us to have consistent widths for every element regardless of the row. The current implementation lets an element's width change depending on the row:

Display name extends past the display name section underneath:
image

Not really a problem and this PR still fixes the overflow issue, so LGTM!

FestiveKyle (approved):

PR #2502, opened 2021-06-23
Organization card overflow

AdamPrins:
Changes width of the org name to 30% for medium (down from 50%). 
Returns to 50% width for large size.

Screen Shot 2021-06-23 at 9 24 58 AM

fixes  #2410

PR #2517, opened 2021-06-25
Ux organization overflow fix

### LINKED ISSUE #2410: Organization card overflows on smaller devices
Organization cards overflow when viewed on smaller devices.

image
image

AdamPrins:
Changes width of the org name to 30% for medium (down from 50%). 
Returns to 50% width for large size.

Screen Shot 2021-06-23 at 9 24 58 AM

fixes  #2410

Replaces #2502

### DIFF
--- a/frontend/src/OrganizationCard.js
+++ b/frontend/src/OrganizationCard.js
@@ -61,7 +61,7 @@ export function OrganizationCard({
         as={!smallDevice ? RouteLink : ''}
         to={`${path}/${slug}`}
       >
-        <Box flexShrink="0" minW="50%" maxW={['100%', '50%']} mb={['2', '0']}>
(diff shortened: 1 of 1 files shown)


FestiveKyle (changes requested):
This causes overflow on some screen sizes:
image

Using a flex or grid here should work well. If using flex, the configuration bars both have a static width and we can give the service count an appropriate width (or flex-basis) based on em units. Since everything right-hand-side of each row will have known widths, everything will be aligned properly in columns. Combining this with appropriate flex-grow's and give every org name the same flex-basis will make get everything aligned and makes sure nothing overflows. DomainCard uses this approach

AdamPrins:
Changed to flex, using DomainCard as a reference.

FestiveKyle (approved):
I'm not sure if we need the progress divs to grow/set their flex-basis, using their auto content width looks pretty good. Besides that, LGTM!

Grow:
image


No-grow:
image

PR #2518, opened 2021-06-25
Remove id query from ORG_DETAILS_PAGE

AdamPrins:
Exposes IDs to accounts with only user permissions in an org
significantly increases query response size
Results were not being used

Replaces  #2505

### DIFF
--- a/frontend/src/graphql/queries.js
+++ b/frontend/src/graphql/queries.js
@@ -617,7 +617,7 @@ export const ORG_DETAILS_PAGE = gql`
           }
         }
       }
-      affiliations {
+      affiliations(first: 1) {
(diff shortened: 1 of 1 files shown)


FestiveKyle (commented):
The api currently requires "first" to be set, so this isn't possible. Keeping the "first" variable without returning any edges results in the user affiliation query not being sent over the network (details page query sets edges to [], then affiliation query thinks it has all the information available in the cache since there are no affiliation edges). Extracting the affiliation count to organization (similar to domain count) is probably the easiest way to fix this without re-writing caching functions or forcing network requests.

FestiveKyle (approved):

PR #2519, opened 2021-06-25
Scanners fix lack of DKIM selectors being assessed as passing

ThomasNickerson:
This PR changes the result processor to fail domains on DKIM if status cannot be assessed for any of their selectors. It also makes the DMARC scans more resilient to missing DMARC records.


Ethanljf (approved):
Looks great! Good stuff🔥

Issue #2414, opened 2021-06-10
Unreachable domains should be distinguished from failing domains


ThomasNickerson:
**Describe the bug**
HTTPS and SSL status is set to fail for domains that cannot be reached.

**To Reproduce**
1. Review results for a domain that can't be navigated to in your browser

**Expected behavior**
Domains that are not reachable should be distinguished from failing domains on their card, and not count towards organization compliance percent calculation.

Issue #2413, opened 2021-06-10
HTTPS bad hostname tag is ambiguous


ThomasNickerson:
**Describe the bug**
The HTTPS bad hostname (https5) tag will be applied to both unreachable domains and domains with certificates that don't match their name.

**To Reproduce**
1. Check results for an unreachable domain
2. Inspect HTTPS scanner [source](https://github.com/canada-ca/tracker/blob/d0b6e8518fd6fe23caa808bf01df4752a7071692/services/scanners/https/scan/https.py#L1275-L1284) and note that his should be applied to domains where there is an issue with the certificate Subject Alternative Name.

**Expected behavior**
Guidance tags should distinguish between the domains that are unreachable and domains that present an invalid Subject Alternative Name

PR #2522, opened 2021-06-25
Unreachable domain handling

Ethanljf:
These changes resolve the following two issues relating to domains that do not resolve:

https://github.com/canada-ca/tracker/issues/2414

https://github.com/canada-ca/tracker/issues/2413

### DIFF
--- a/services/scanners/results/result_processor.py
+++ b/services/scanners/results/result_processor.py
@@ -312,7 +312,7 @@ def process_dns(results, domain_key, uuid, db):
     timestamp = str(datetime.datetime.utcnow())
     tags = {"dmarc": [], "dkim": {}, "spf": []}
 
-    if results["dkim"].get("error") == "missing":
+    if not results["dkim"].get("error"):
         for selector in results["dkim"].keys():
             tags["dkim"][selector] = []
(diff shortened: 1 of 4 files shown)


services/scanners/results/result_processor.py
```diff
@@ -843,9 +851,7 @@ def process_dns(results, domain_key, uuid, db):
                 elif all(i in ["dkim7", "dkim8"] for i in tags["dkim"][selector]):
                     dkim_statuses.append("pass")
 
-            if len(dkim_statuses) == 0:
-                dkim_status = "fail"
-            elif any(i == "fail" for i in dkim_statuses):
+            if any(i == "fail" for i in dkim_statuses):
                 dkim_status = "fail"
             else:
                 dkim_status = "pass"
```
ThomasNickerson, line 857: I think this reverts the fix to missing selectors passing, because if there are no selectors dkim_statuses should be empty instead of containing any "fail"s. For now, if we cant assess status for any selectors that should be considered a fail.

ThomasNickerson (changes requested):

ThomasNickerson (approved):

Issue #2453, opened 2021-06-16
Core Service Needs to be Restructured 


nsdeschenes:
The core service needs to be judo chopped by the hand of separation of concerns. 

The core service is currently handling updating the guidance tags in the database, calculating scan summaries for individual organizations, and the system as a whole. 

This causes issues when the core service is being ran before the scans for that day, the scan summary shown in the UI does not match that of the actual scan data, it is actually two days older. This may also have a side effect with MAF calculations if they are being computed on old data.

The core service needs to be broken up into two different services one for updating guidance tags prior to the scans occurring, and one for calculating the summaries post scanning.

PR #2528, opened 2021-06-25
Core service restructure

Ethanljf:
These changes address concerns raised in issue https://github.com/canada-ca/tracker/issues/2453

This removes the core service and splits the operations into two services. "Guidance" will update guidance and criteria and will run before scanning, "summaries" will run after scanning to ensure that summaries are calculated correctly.


app/bases/knative/config/summaries.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: summaries
+  namespace: scanners
+spec:
+  schedule: "0 5 * * *"
```
nsdeschenes, line 7: How can we guarantee that this is always after the scans have been ran?

services/summaries/summaries.py
```diff
@@ -0,0 +1,194 @@
...
+def update_scan_summaries(host=DB_HOST, name=DB_NAME, user=DB_USER, password=DB_PASS, port=DB_PORT):
+    logging.info(f"Updating scan summaries...")
+
+    # Establish DB connection
+    connection_string = f"http://{host}:{port}"
+    client = ArangoClient(hosts=connection_string)
+    db = client.db(name, username=user, password=password)
+
+    if not db.has_collection("scanSummaries"):
+        db.create_collection(
```
nsdeschenes, line 34: This needs to be removed for the database initilization script to be the so it does not override any settings that we could possibly change.

services/summaries/summaries.py
```diff
@@ -0,0 +1,194 @@
...
+def update_chart_summaries(host=DB_HOST, name=DB_NAME, user=DB_USER, password=DB_PASS, port=DB_PORT):
+    logging.info(f"Updating chart summaries...")
+
+    # Establish DB connection
+    connection_string = f"http://{host}:{port}"
+    client = ArangoClient(hosts=connection_string)
+    db = client.db(name, username=user, password=password)
+
+    if not db.has_collection("chartSummaries"):
+        db.create_collection(
```
nsdeschenes, line 85: This needs to be removed for the database initilization script to be the so it does not override any settings that we could possibly change.

nsdeschenes (changes requested):
The `update_guidance` function also needs the database creation functions removed, same reasons as above.

ThomasNickerson (changes requested):
We also need to update the flux image policy and flux image registry to reflect the split, probably splitting it up the same way everything else was split makes the most sense.

Everything else looks great!

https://github.com/canada-ca/tracker/blob/master/deploy/bases/core-registry.yaml
https://github.com/canada-ca/tracker/blob/master/deploy/gke/core-image-policy.yaml

app/bases/knative/config/summaries.yaml
Ethanljf: Updated to reflect the schedule agreed upon in our discussion

ThomasNickerson (approved):

PR #2532, opened 2021-06-28
Changing phone number will update the TFA settings

### LINKED ISSUE #2335: TFA Options Not Updating After Adding Phone Number
**Describe the bug**
After adding my phone number to my account, I went to set the TFA method to use text, however the option is not showing up.

**To Reproduce**
1. Go to 'Account Settings' Page
2. Add Phone Number
3. Click on the 'Two-Factor Authentication' Drop Down

**Expected behavior**
After adding the users phone number the drop down should include the option for the users to set their phone number.

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Brave Browser 
 - Version: 91.1.25.68

AdamPrins:
Adding or removing a phone number will now update the TFA checks and drop down menu.

fixes  #2335

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -22,7 +22,7 @@ import { useUserVar } from './UserState'
 const PageNotFound = lazy(() => import('./PageNotFound'))
 const CreateUserPage = lazy(() => import('./CreateUserPage'))
 const DomainsPage = lazy(() => import('./DomainsPage'))
-const UserPage = lazy(() => import('./UserPage'))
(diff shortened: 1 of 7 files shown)


AdamPrins:
The wrapper is required because the useState has to be before any returns, 
But also needs to be after the graphql query and the loading check.

FestiveKyle (changes requested):
This fix can be implemented by returning the new phone validated and TFA send method states in the mutations.
#2533

AdamPrins:
Fix implemented by #2533

PR #2547, opened 2021-06-29
Switch Refresh Tokens to Cookies

nsdeschenes:
This PR re-works the original implementation of refresh tokens by utilizing `httpOnly` cookies.
The changes includes:
- Removal of the `refreshTokens` field from the `AuthResult` gql object
- Addition of new `signOut` mutation that clears the cookie
- The following mutations now set the `refresh_token` cookie: `authenticate`, `signIn` (non-tfa), and `signUp`

Example `signOut`:
```graphql
mutation {
  signOut (input: {}) {
    status
  }
}
```

Example `refreshTokens`:
```graphql
mutation {
  refreshTokens(input: {}) {
    result {
      ... on AuthResult {
        authToken
      }
    }
  }
}
```


api-js/src/user/mutations/refresh-tokens.js
```diff
@@ -141,24 +145,31 @@ export const refreshTokens = new mutationWithClientMutationId({
...
       expPeriod: 168,
       secret: String(REFRESH_KEY),
     })
 
-    console.info(`User: ${user._key} successfully refreshed their tokens.`)
+    response.cookie('refresh_token', newRefreshToken, {
+      maxAge: REFRESH_TOKEN_EXPIRY * 60 * 24 * 60 * 1000,
+      httpOnly: true,
+      secure: false,
+      sameSite: true,
```
FestiveKyle, line 167: Is sameSite: true a valid value?

lcampbell2: sameSite: true is the same as sameSite: 'strict'. It used to binary but then they added 'lax' as an in-between

Ethanljf (approved):

lcampbell2 (approved):

FestiveKyle (approved):

PR #2551, opened 2021-06-29
Api domain query optimizations

ThomasNickerson:
This PR removes unneeded calls to `DOCUMENT` in AQL query strings that were found to have a negative performance impact. Affected operations (domain retrieval by user or by org) should see a large performance improvement.

For example, here is a comparison of the performance of a query getting 100 domains after a particular ID, sorted by HTTPS status:

Before:
before

After:
after (2)


sleepycat (approved):
:1st_place_medal: 
This is awesome!

nsdeschenes (approved):

PR #2571, opened 2021-07-02
Disable query tracing by default

ThomasNickerson:
Set `TRACING_ENABLED` to be false by default in makefile


sleepycat (approved):
:brain:

PR #2573, opened 2021-07-02
Api optimize guidance tag queries

ThomasNickerson:
This PR refactors the guidance tag connection loaders similarly to #2551, resulting in better performance.


sleepycat (approved):
:star_struck:

nsdeschenes (approved):

PR #2587, opened 2021-07-07
Allow Users to Leave Organizations

nsdeschenes:
This PR introduces a new PR that allows users to leave to leave an organization. If the user is the "owner" of that organization i.e. the one who created it, then the organization, affiliations will be removed, alongside any domain if the given org was the last one claiming that domain the scan info, and possibly dmarc summary data will be removed alongside.

If the user is not the "owner" the user will just have their affiliation removed.

Verified orgs are not affected by this because when orgs are transitioned to the verified state via the `verifyOrganization` mutation all the affiliations have `owner: false` set.


api-js/src/affiliation/mutations/leave-organization.js
```diff
@@ -0,0 +1,358 @@
...
+
+    const user = await userRequired()
+
+    verifiedRequired({ user })
+
+    const org = await loadOrgByKey.load(orgKey)
+
+    if (typeof org === 'undefined') {
+      console.warn(
+        `User: ${user._key} attempted to leave org: ${orgKey}, however it's undefined.`,
```
Ethanljf, line 46: I would probably say  "User ${user._key} attempted to leave undefined organization: ${orgKey}" just for the sake of clarity

Ethanljf (approved):

PR #2593, opened 2021-07-07
Add phase and status to dmarc scan subscription

FestiveKyle:
As described

### DIFF
--- a/api-js/src/email-scan/objects/dkim-sub.js
+++ b/api-js/src/email-scan/objects/dkim-sub.js
@@ -1,6 +1,7 @@
 import { GraphQLObjectType, GraphQLList, GraphQLID } from 'graphql'
 
 import { domainType } from '../../domain/objects'
+import { StatusEnum } from '../../enums/status'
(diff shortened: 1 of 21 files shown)


nsdeschenes:
The issue with DMARC Phase is that it's directly not related to only the dmarc related scans, it is related with DKIM, and SPF as well, so this is something we're going to have to discuss.

ThomasNickerson (approved):

api-js/src/email-scan/objects/dkim-sub.js
```diff
@@ -1,6 +1,7 @@
 import { GraphQLObjectType, GraphQLList, GraphQLID } from 'graphql'
 
 import { domainType } from '../../domain/objects'
+import { StatusEnum } from '../../enums/status'
```
nsdeschenes, line 4: Can you switch this import to, please and thank you
```js
import { StatusEnum } from '../../enums'
```

nsdeschenes (changes requested):

nsdeschenes (approved):

PR #2594, opened 2021-07-08
Switch traversals in domain connection loaders to breadth first

ThomasNickerson:
This PR switches several traversals to breadth first as it was discovered that this drastically improved execution time in some cases.

Before:
dfs

After:
bfs


sleepycat (approved):
:star_struck:

PR #2605, opened 2021-07-09
Use breadth first traversal in results loaders

ThomasNickerson:
This PR improves performance when getting scan results by switching a traversal in the results queries to breadth first.

Before (profiling done on HTTPS results):
dfs_https_results

After:
bfs_https_results


sleepycat (approved):
:fire:

Issue #1996, opened 2021-03-30
Creating French Translations for Guidance Tags


nsdeschenes:
**Is your feature request related to a problem? Please describe.**
This issue is related to guidance tags not being translated to french, when they should be.

**Describe the solution you'd like**
I would like to see the guidance tag json files to be updated to have a structure like the following:
```json
{
  "dkim1": {
    "en": {
      "tagName": "DKIM-GC",
      "guidance": "Government of Canada domains subject to TBS guidelines",
      "refLinksGuide": [
          {
              "description": "IT PIN"
          }
      ],
      "refLinksTechnical": [""]
    },
    "fr": {
      "tagName": "DKIM-GC",
      "guidance": "Les domaines du gouvernement du Canada sont soumis aux directives du SCT.",
      "refLinksGuide": [
          {
              "description": "IT PIN"
          }
      ],
      "refLinksTechnical": [""]
    }
  }
}
```

**Additional context**
This would require an update to the core service to insert the guidance tags differently and as well as an update to the API so that it is able to handle the translation.

PR #2615, opened 2021-07-12
Update guidance service for translations on guidance tags

ThomasNickerson:
This PR updates guidance tags and the guidance service to implement the format outlined in #1996.


ThomasNickerson:
Superseded by #2712

PR #2617, opened 2021-07-12
Add one time scans to the frontend

FestiveKyle:
Adds individual scan cards for one time scans. These components should be refactored for abstract use with ScanCard in the future.

image

image

image


frontend/src/client.js
```diff
@@ -65,7 +100,10 @@ export const currentUserVar = makeVar({
 })
 
 const httpLink = createHttpLink({
-  uri: '/graphql',
+  uri:
+    process.env.NODE_ENV === 'production'
+      ? 'https://tracker.alpha.canada.ca/graphql'
+      : '/graphql',
```
sleepycat, line 106: Tracker is going to be deployed at https://tracker.canada.ca and is currently running at https://staging.tracker.canada.ca and https://tracker.ops.cyber.gc.ca as well as our GCP environment at https://tracker.alpha.canada.ca.
I think leaving this at `/graphql` is probably a good idea.

frontend/src/client.js
```diff
@@ -80,7 +118,39 @@ const headersLink = setContext((_, { headers }) => {
   }
 })
 
+const httpLinkWithHeaders = headersLink.concat(httpLink)
+
+const wsLink = new WebSocketLink({
+  uri:
+    process.env.NODE_ENV === 'production'
+      ? 'wss://tracker.alpha.canada.ca/graphql'
+      : 'ws:localhost:3000/graphql',
```
sleepycat, line 127: Same as above.

sleepycat (changes requested):
Looks good but we're baking some assumptions about urls in there! Update those and we're good!

sleepycat (approved):
Nice! :star_struck:

lcampbell2 (approved):

PR #2626, opened 2021-07-15
Unverify Usernames When Updated

nsdeschenes:
This PR updates the `updateUserProfile` mutation and will set the users `emailVerified` value to `false`, only when the user updates their username. In turn a notify email is sent so the user can verify their new username.


Ethanljf (approved):

sleepycat (approved):
:+1:

PR #2644, opened 2021-07-19
Fix Lint warnings

AdamPrins:


frontend/src/__tests__/useDebouncedFunction.test.js
```diff
@@ -8,7 +8,7 @@ const UseDebouncedFunctionExample = () => {
 
   const memoizedSetDebouncedCountCallback = useCallback(() => {
     act(() => setDebouncedCount(debouncedCount + 1))
-  }, [count])
+  }, [debouncedCount])
```
FestiveKyle, line 11: We need to use `count` as the only variable in the dependency array here. Including debouncedCount here will cause an infinite loop. We can see this by including this case at the bottom of the test:
```
// ensure no more increments
jest.advanceTimersByTime(1000)
expect(debouncedCount).toHaveValue('2')
```
I'll put up a PR with some more cases and a different implementation for the useCallback function.

Some info on how useDebouncedFunction works in case you want to know more:
useDebouncedFunction runs a function after a timeout. If the function handed to useDebouncedFunction changes while the timeout hasn't finished yet, a new timeout instance will start (with the new function given to useDebouncedFunction), and the uncalled function is thrown away and won't be called.
useCallback only re-creates a function after the first render and any time something in the dependency array changes. The functionality we want is it to be re-created every time `count` is changed. This lets us make sure that some call are being discarded.

FestiveKyle: PR with some changes:
https://github.com/canada-ca/tracker/pull/2648

PR #2646, opened 2021-07-19
improve AdminPage formatting

AdamPrins:
adds gap between org select and create org on small screens

adds gap between buttons and UserCard on small screens

fixes spacing of delete/edit org on small screens

fixes size of verified check

better uses space on large screens by making Domain information one line

image


FestiveKyle (approved):
🤩

PR #2649, opened 2021-07-19
update ambiguous arrays to objects

AdamPrins:
change from array notation to object notation:
```
fontSize={['2xl', '2xl', '2xl', '3xl', '4xl']}
fontSize={{ base: '2xl', lg: '3xl', xl: '4xl' }}
```

This means that breakpoints are much clearer

also fixes points that were given two items in the array, where the breakpoint should have been medium, but was actually small.


FestiveKyle (approved):
Good stuff! 🔥

PR #2650, opened 2021-07-19
Fix RQ workers on OTS queues

ThomasNickerson:
This PR fixes OTS priority queues by changing environment variable names that were causing unintended RQ behavior.


Ethanljf (approved):
Awesome stuff! This looks great!

PR #2658, opened 2021-07-20
update Layout.js

AdamPrins:
Simplify Layout

Layout only used in PhaseBanner, TopBanner, Footer

use LayerStyle as the one source for maxw

simplify footer

Replaces  #2645


AdamPrins:
Merge conflicts fixed on #2687

PR #2659, opened 2021-07-20
Fix flux for queues

ThomasNickerson:
This PR removes references to deleted queue overlays and fixes image paths for queues.


sleepycat (approved):
:+1:

PR #2660, opened 2021-07-20
Remove unused js files

AdamPrins:
If we want to use TwoFactorNotificationBar, we can restore it from the history.
I think it should be removed now because it is not being implemented.


sleepycat (approved):
:broom: :house: :exclamation:

PR #2661, opened 2021-07-20
Replace collapse and states, with Accordion

AdamPrins:
Uses built in Chakra Accordion

Buttons now show state of panel with the Chevron

Categories are now clearly buttons that collapse

Ownership is shown by small x padding

Screen Shot 2021-07-20 at 11 47 04 AM
(Button is orange due to hover)

mobile
Screen Shot 2021-07-20 at 11 49 32 AM


lcampbell2 (approved):
LGTM :fire:

PR #2696, opened 2021-07-23
Make `No ...` messages the same size as loading messages

### LINKED ISSUE #2628: Invite box jumps while searching
The invite box is rerendered in a way that makes the page jump.
 
jumpy-input

AdamPrins:
Should fix #2628

but unable to properly test because local environment always displays some users/domains/orgs


sleepycat (approved):
:+1:

PR #2697, opened 2021-07-23
Add labels it IconButtons

### LINKED ISSUE #2636: Some buttons labels need to be more verbose
Buttons such as the "Edit" and "Save" buttons on the Account Settings page need more context for accessibility. This can be done by setting the `aria-label` property to something more verbose (such as `aria-label="Edit Display Name"`.

AdamPrins:
All buttons that have no text now have a proper label

fixes  #2636

If you want all buttons, even with text, to have labels, I can add onto this pull request.


sleepycat (approved):
:+1:

PR #2705, opened 2021-07-26
Assessibility pass

AdamPrins:
Hides decorative icons

adds labels to meaningful icons

and text labels to inputs 

adds tabbability to some elements

some things that still need fixes
contrast colours
table headers are tabbable, but don't change sort when keys are pressed
Dmarc report graph needs label/summary

### DIFF
--- a/frontend/src/__tests__/AdminDomains.test.js
+++ b/frontend/src/__tests__/AdminDomains.test.js
@@ -254,7 +254,7 @@ describe('<AdminDomains />', () => {
       )
 
       const domainUrlInput = await findByRole('textbox', {
-        label: 'Search by Domain URL',
+        name: 'Search by Domain URL',
       })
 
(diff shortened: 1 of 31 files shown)


FestiveKyle (changes requested):
There's a few `for` props that should be switched to `htmlFor`

frontend/src/__tests__/AdminDomains.test.js
```diff
@@ -254,7 +254,7 @@ describe('<AdminDomains />', () => {
       )
 
       const domainUrlInput = await findByRole('textbox', {
-        name: 'Domain URL',
+        label: 'Search by Domain URL',
```
FestiveKyle, line 257: This should still use the `name` option. `label` isn't an option, so findByRole just matches to the textbox anyway since it's the only textbox there

FestiveKyle (changes requested):

FestiveKyle (approved):

PR #2712, opened 2021-07-27
Look Look, We Finally Did It!

nsdeschenes:
Yes folks it is finally here, translated guidance tags, three branches later we are finally here.

So this PR updates the various guidance tag JSON files updating them to have their appropriate translations ensuring. The guidance service that updates the DB with the guidance tag info has also been updated to handle these changes. The API in turn has had all the various guidance tag loaders updated to handle the translations determined upon the requesting users language.

### DIFF
--- a/api-js/jest.config.js
+++ b/api-js/jest.config.js
@@ -11,5 +11,5 @@ module.exports = {
     'index.js',
     'env.js',
   ],
-  testTimeout: 50000,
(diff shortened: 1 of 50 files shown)


ThomasNickerson (approved):
LGTM!

Ethanljf (approved):

PR #2752, opened 2021-07-30
Manage processes in queue containers with Supervisord

ThomasNickerson:
RQ docs state that workers should be managed by a process manager such as [supervisord](http://supervisord.org/index.html#) or systemd in production systems; this PR configures Tracker's queues to have all their component processes (Redis, RQ workers, gunicorn) managed by supervisord. This will improve reliability by enabling automatic restart at the level of individual processes, as Kubernetes is only able to monitor and orchestrate at the container/pod level.

A positive side effect is that Redis logging output should be collected by GKE now. Previously this was not possible as Redis will not log to stdout when daemonized, but supervisord is able to redirect its logging to stdout.

Note: number of workers can be configured by changing `numprocs=4` to the desired number for each worker type in supervisord.conf.


Ethanljf (approved):
Looks great! Looking forward to these changes!

PR #2783, opened 2021-08-05
fix landing page image

AdamPrins:


AdamPrins:
Landing page text reverted, image renders fine again.

PR #2784, opened 2021-08-05
Remove manual json encoding for posts

FestiveKyle:
As described


Ethanljf (approved):
This is awesome! Great catch👍

PR #2831, opened 2021-08-13
Organize frontend by route

AdamPrins:
Rough pass to get everything into a reasonable spot.

Some folders might be better folded in together.


lcampbell2 (changes requested):
The tests should be structured as the api has done where there is a `__tests__` directory in each route folder. So instead of `AdminPage.test.js` being located at `/__tests__/admin/AdminPage.test.js`, it should be `/admin/__tests__/AdminPage.test.js`

lcampbell2 (approved):
:100: :100: :100:

PR #2837, opened 2021-08-16
Extract domain card status badge

AdamPrins:
Removes code duplication found in DomainCard


FestiveKyle (approved):
Great stuff, way more clean 🔥

PR #2846, opened 2021-08-18
Remove Python version req from pipfile

ThomasNickerson:
Setup.py already specifies `python_requires=">=3.6"` and pipenv does not support multiple Python version requirements, or ranges, so this section should be removed from the pipfile.


sleepycat (approved):
:+1:

PR #2852, opened 2021-08-19
DMARC Report Service Reworking

nsdeschenes:
This PR contains changes to the way that the DMARC Report service functions.

Changes:
- Fix issue with old date values not being removed properly
- Dates are now compared between Cosmos and Arango
- Ownerships are only created, removed, or updated when need be lowering run times
- Increased overall performance
- Introduced use of transactions for any data modification in Arango
- Removed unused functions
- 100% Test coverage across the board


sleepycat (approved):
:+1:  Nice!

PR #2938, opened 2021-09-09
Fix gc.ca Domain Scalar Issues

### LINKED ISSUE #2841: Domain "gc.ca" breaks domain pages
**Describe the bug**
If `gc.ca` is loaded by a query, it returns an error citing that it is not a valid domain

**To Reproduce**
1. Go to the `Government of Canada` organization on the Organizations or Admin page
2. Attempt to view domains
3. See error

**Expected behavior**
If domain is invalid, it should not be able to be added

**Screenshots**
Admin Page:
Screenshot from 2021-08-17 11-06-50
Organizations Page:
Screenshot from 2021-08-17 11-07-01
Attempting to load on Domains page results in eternal loading:
Screenshot from 2021-08-17 11-08-50

nsdeschenes:
This PR aims to fix the issues with the `gc.ca` domain. The issue that is being brought up is that `gc.ca` is itself a tld on the public suffix list [package](https://www.npmjs.com/package/psl) that we were originally using. This PR removes the `psl` package entirely and switches to using regex to validate the domains.

Closes #2841

### DIFF
--- a/api-js/package-lock.json
+++ b/api-js/package-lock.json
@@ -37,7 +37,6 @@
         "jsonwebtoken": "^8.5.1",
         "moment": "^2.29.1",
         "notifications-node-client": "^5.1.0",
-        "psl": "^1.8.0",
         "subscriptions-transport-ws": "^0.9.19",
         "url-slug": "^3.0.2",
         "uuid": "^8.3.2",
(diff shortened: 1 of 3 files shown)


sleepycat (approved):
:+1:

PR #3005, opened 2021-09-22
Expand Domain Status Fields

nsdeschenes:
This PR introduces new status fields for the `domainStatus` GQL object, providing more detailed information for the domain card list thing. It also includes changes to the result processor that will set default `fail` values until the new scanners are introduced.


FestiveKyle (approved):
👍

PR #3075, opened 2021-10-04
Add new fields to domain summary cards

FestiveKyle:
Adds "Ciphers", "Curves", "HSTS", and "Protocols" fields to the domain summary cards.

image


maxneuvians:
Moved to #3151 because of the legacy cloudbuild process that fails the tests

PR #3185, opened 2021-11-09
Remove usage of ttl cron jobs

FestiveKyle:
Increase job ttl to 6hr from 100sec. Remove usage of ttlSecondsAfterFinished from CronJobs. Using CronJob's default `successfulJobsHistoryLimit` and `failedJobsHistoryLimit` options, 3 successful and 1 failed job will be allowed for history.


sleepycat (approved):
:+1:

PR #3314, opened 2021-12-14
feat: log4shell dns capture

maxneuvians:
This PR converts the log4shell processor from an LDAP server to a DNS server that will capture queries made against a specific domain pattern. The pattern can be defined through the `TOKEN` and `DOMAIN` env variables but default to the following:

```
const domain = DOMAIN || 'log4shell.tracker.alpha.canada.ca'
const token = TOKEN || 't'
```

Assuming the nameserver is handling requests for  `log4shell.tracker.alpha.canada.ca` 

it would log requests like: 

`foo.t.log4shell.tracker.alpha.canada.ca`.

But not for 

`foo.bar`

or 

`foo.log4shell.tracker.alpha.canada.ca`. (which is missing the `token` which in this case is `t`).

The reason for using the token is to not log errant requests on the domain name, for example `smtp.log4shell.tracker.alpha.canada.ca`.

Code can be tested with the docker image with UDP, ex:

```
docker run -p 5053:53/udp {IMAGE_NAME}
dig A example.com.t.log4shell.tracker.alpha.canada.ca @127.0.0.1 -p5053
```

or TCP

```
docker run -p 5053:53 {IMAGE_NAME}
dig A example.com.t.log4shell.tracker.alpha.canada.ca +tcp @127.0.0.1 -p5053
```


sleepycat (approved):
:star_struck:

PR #3323, opened 2021-12-14
Start Hashing Domains

nsdeschenes:
This PR modifies the API `createDomain` function to include a new salted hashing function that will create a new field in the db for the hashed string. As well the `domain-loader` script has been updated to include this hashing during its loading process.


sleepycat (approved):
:star_struck:

sleepycat (approved):
:raised_hands:

PR #3500, opened 2022-02-28
Update domain list orderby options

lcampbell2:
Update orderBy enums and selection for domain lists to match the status' displayed.


FestiveKyle (approved):
LGTM  :+1:

PR #3638, opened 2022-04-28
Add Tabs to Routing

### LINKED ISSUE #3283: Directly navigating to an organization which doesn't exists gives no error
**Describe the bug**
When navigating to an organization page which doesn't exist, an organization page is shown but with empty areas.

**Expected behavior**
An error should be displayed to the user stating that either the organization does not exist or they don't have access to it.

**Screenshots**
image

### LINKED ISSUE #3468: Add clarity to why not all domains show in HTTPS chart
Currently the charts for HTTPS summaries show only a subset of all domains as it only includes web-hosting domains. We should clarify this - possibly with local text explaining so or by adding another "info" dataset to the chart.

- [ ] Wording needed to advise users that the results may not show all domains. (only webhosting domains....example).

image

lcampbell2:
- add tab names to routing on some pages to persist in history traversal
- update UserPage tests
- Clean up domain/org cards
- display error when navigating to org that doesn't exist (closes #3283)
- adds domain total and text to Summary Donuts (closes #3468)

### DIFF
--- a/frontend/src/admin/__tests__/UserListModal.test.js
+++ b/frontend/src/admin/__tests__/UserListModal.test.js
@@ -277,7 +277,7 @@ describe('<UserListModal />', () => {
     })
     describe('admin has "SUPER_ADMIN" privileges', () => {
       describe('admin is updating user with "USER" privileges', () => {
-        it('admin can change user role to "ADMIN"', async () => {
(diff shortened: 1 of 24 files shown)


FestiveKyle (approved):
LGTM

PR #3709, opened 2022-05-24
Frontend unused deps

lcampbell2:
- removes unused deps in the frontend
- updates webpack deps to remove security issues


sleepycat (approved):
:exploding_head: That's surprising! Nice work!

PR #3925, opened 2022-08-17
myTracker View

### LINKED ISSUE #3892: Private domain view: myTracker
In the past there was the idea for users to create their own "sandbox" orgs where they can manage and track domains. Building on that idea and the desire to slightly gamify getting domains compliant, a new view is to be created where users can "favourite" domains across the app and track their web and email status'. The view will give similar information as the Organization Details pages.

lcampbell2:
Users can now tag domains as favourites and have a personal "org" view of these tagged domains, including the summary cards, dmarc phase chart, and domain list.

New favourite button on domain cards:
Screenshot from 2022-08-17 11-27-42

Each user can now be a personal "organization":
Screenshot from 2022-08-17 11-27-12

Unfavourite domains just as easily as you favourite them:
Screenshot from 2022-08-17 11-27-28

closes #3892


frontend/src/domains/DomainCard.js
```diff
@@ -127,13 +188,40 @@ export function DomainCard({ url, status, hasDMARCReport, tags, ...rest }) {
...
+                      await favouriteDomain({ variables: { domainId: id } })
+                    }}
+                    variant="primary"
+                    aria-label={`favourite ${url}`}
+                    icon={<StarIcon />}
+                  />
+                ))}
+            </ABTestVariant>
+          </ABTestingWrapper>
+        </Stack>
```
FestiveKyle, line 217: It would be clean if we could show the `favourite` status on the normal "Domains" page, while also allowing the user to unfav on the "Domains" page. How feasible is that?

api/database.json
```diff
@@ -203,6 +203,15 @@
...
     },
+    {
+      "type": "edgecollection",
+      "name": "favourites",
+      "options": {
+        "replicationFactor": 3,
+        "writeConcern": 1,
+        "numberOfShards": 6
+      }
+    },
```
FestiveKyle, line 214: I think these changes might have to be added in `/k8s/apps/bases/api/database-configmap.yaml` as well, so the `database-migration` init container will create this collection. We should probably look moving the `database.json` into a good spot where it can be used wherever required.

FestiveKyle (changes requested):
Looks great :fire:  Just 1 bug with the mock schema and a couple things to note.

frontend/src/domains/DomainCard.js
lcampbell2: Yeah, I agree. I'm thinking we add that in later, and merge in what we have now.

api/database.json
lcampbell2: Good catch. Fixed :+1:

frontend/mocking/faked_schema.js
```diff
@@ -3057,6 +3114,33 @@ export const getTypeNames = () => gql`
...
+  }
+
+  #
+  input InputTagLabel {
+    # The English translation of the label.
+    en: String
+
+    # The French translation of the label.
+    fr: String
+  }
```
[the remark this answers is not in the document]
lcampbell2: Merge conflicts love messing things up. Fixed :+1:

FestiveKyle (approved):

PR #4046, opened 2022-10-12
Update documentation1

h701h:
When this string is used it mostly points to TBS documentation. Therefore, removing the reference to CCCS.


lcampbell2 (changes requested):
The reference to CCCS should also be removed in the msgid, and from the actual code in GuidanceTagDetails.js:18

lcampbell2 (approved):

PR #4074, opened 2022-10-24
Add composition to scan cards

FestiveKyle:
As described


frontend/src/app/App.js
```diff
@@ -26,6 +26,7 @@ import { IS_LOGIN_REQUIRED } from '../graphql/queries'
 import { useLingui } from '@lingui/react'
 import { ABTestingWrapper } from './ABTestWrapper'
 import { ABTestVariant } from './ABTestVariant'
+import GuidancePage from '../guidance/GuidancePage'
```
lcampbell2, line 29: should lazy load like the other pages

lcampbell2 (changes requested):
Love how much smaller the components are!

PR #4317, opened 2023-01-30
Fix register terms and conditions

### LINKED ISSUE #4188: Bug: I agree to all Terms, Privacy Policy & Code of Conduct Guidelines
While registering an account the "I agree to all Terms, Privacy Policy & Code of Conduct Guidelines" is not clickable.

lcampbell2:
Add link to terms agreement when creating account

closes #4188


FestiveKyle (approved):
I think some distinction between the link and the previous text would be helpful - bold, underline, different colour. Something like that.

:+1:

PR #4508, opened 2023-05-04
Add script to downgrade list of affiliations to the "USER" level

### LINKED ISSUE #4471: Prepare for downgrade of user rights
- [x] Have a backup of all roles and users
- [ ] Script to flip them all
- [ ] This is an auditable event 
- [ ] Admins -> Users

FestiveKyle:
As described.

Closes #4471


scripts/downgrade-users/example.csv
```diff
@@ -0,0 +1,3 @@
+userId,displayName,userName,totalCount,affiliationId,permission,organizationId,acronym,name,slug
```
lcampbell2, line 1: It looks like in the script we're only using the orgId and userName values. Are the other headers necessary?

FestiveKyle: Nope, not required for the script. The other stuff is there since that's the info we took from the export and it might be easier to vet with the extra columns. We can remove from example.csv no problem

PR #4531, opened 2023-05-17
Update `make cluster` command

sleepycat:
This commit updates the `make cluster` command in a few ways:
* Removed `beta` since no options require it
* replaced project name with project variable
* Hardcode k8s version to 1.24 (since higher requires config changes)
* Set release channel to stable (rapid is... rapid.)
* Uses a least privilege service account

The makefile now includes a command for making that least privilege service account as well.

The result is that cluster can be created, and
existing config can be applied without errors/warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,16 @@ endef
 
 .PHONY: cluster
 cluster:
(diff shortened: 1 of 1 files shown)


lcampbell2 (approved):
LGTM :smile_cat:

PR #4597, opened 2023-06-20
Remove claimed domains from unclaimed org

### LINKED ISSUE #4582: Periodically remove claimed domains from Unclaimed org
Create a weekly cronjob to remove domains from the Unclaimed org if another (verified?) org has claimed it

lcampbell2:
A cronjob service to periodically remove domains from the "Unclaimed" org if they have been added to another org.

closes #4582

### DIFF
--- a/services/domain-cleanup/src/remove-nxdomain-service.js
+++ b/services/domain-cleanup/src/remove-nxdomain-service.js
@@ -8,7 +8,7 @@ const removeNXDomainService = async ({ query, log }) => {
     try {
       await (
         await query`
-          WITH ownership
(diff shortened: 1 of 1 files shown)


services/domain-cleanup/src/unclaimed-cleanup-service.js
```diff
@@ -0,0 +1,27 @@
+const { UNCLAIMED_ORG_ID } = process.env
```
FestiveKyle, line 1: Thoughts on searching for the org at runtime instead of setting the ID?

FestiveKyle (approved):

PR #4599, opened 2023-06-22
Make view results button visible on DomainCard for unauth'd users 

lcampbell2:
As described, to ensure users know this is an option they can receive upon login and affiliation

### DIFF
--- a/frontend/src/domains/DomainCard.js
+++ b/frontend/src/domains/DomainCard.js
@@ -204,21 +204,19 @@ export function DomainCard({
           justifyContent="center"
           ml={{ base: 0, md: '4' }}
         >
-          {isEmailValidated() && userHasPermission && (
-            <Button
-              variant="primary"
-              as={RouteLink}
-              to={{
-                pathname: `/domains/${url}`,
-                state: { from: location.pathname },
-              }}
-              px="10"
-            >
-              <Text whiteSpace="noWrap">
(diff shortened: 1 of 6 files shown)


frontend/src/guidance/GuidancePage.js
```diff
@@ -59,8 +59,22 @@ function GuidancePage() {
     dmarcPhase,
     rcode,
     status,
+    userHasPermission,
   } = data.findDomainByDomain
 
+  if (userHasPermission) {
```
FestiveKyle, line 65: Should this check be flipped? `!userHasPermission`?

frontend/src/domains/DomainCard.js
```diff
@@ -204,21 +204,19 @@ export function DomainCard({
...
+            to={{
+              pathname: isLoggedIn() ? `/domains/${url}` : '/sign-in',
+              state: { from: location.pathname },
+            }}
+            px="10"
+          >
+            <Text whiteSpace="noWrap">
+              <Trans>View Results</Trans>
+            </Text>
+          </Button>
```
FestiveKyle, line 219: Thoughts on disabling button if `userHasPermission === false`, possibly also changing text to some form of `Not Affiliated` ?

FestiveKyle (changes requested):

FestiveKyle (approved):

Issue #4573, opened 2023-06-05
Load Faster


h701h:
The /domains is loading slow. It needs to be more responsive. 

Options?

1. TBD
2.

PR #4677, opened 2023-07-19
Change database from cluster to single server

FestiveKyle:
The PR changes the ArangoDeployment from `Cluster` mode to `Single` mode. This creates major speedups as it allows for better indexing and removes the inter-server communication required for our many edge-join queries.

Also changes the `fast` disk type on Azure to `PremiumV2_LRS` as it is much more suited to this job (cheaper and faster).

PR effort for #4573.


lcampbell2 (approved):
:+1::+1:

PR #4734, opened 2023-08-03
Domain Discovery Scanner

lcampbell2:
- uses [Findomain](https://github.com/Findomain/Findomain) for subdomain enumeration
- takes base domain and orgId as input
- enumerates based on base domain
- adds new domains to org
- dispatches new domains for scanning


scanners/domain-discovery/domain_discovery.py
```diff
@@ -0,0 +1,193 @@
...
+                )
+                domains_to_scan.append(domainInsert)
+            else:
+                domainInsert = checkDomain
+            db.collection("claims").insert(
+                {
+                    "_from": orgId,
+                    "_to": domainInsert["_id"],
+                    "hidden": False,
+                    "tags": [],
```
FestiveKyle, line 79: Might be worth adding a tag or some other flag for this since it won't be in audit logs.

FestiveKyle (approved):
:+1:

PR #4750, opened 2023-08-17
Only send footprint changes

### LINKED ISSUE #4472: Send notification when an orgs settings or digital footprint changes
- [ ] **Admins and Org Owners** get email notifications when events are created in audit logs - Nightly Batch Job
  - [x] All user CRUD activity
  - [x] All domain CRUD activity
  - [x] All domain tag activity
  - [x] Only send emails if new activity in logs is found. 
  - [x] All exports of orgs data (though this not the big-big export)
  - [x] Nice to have: Disable feature to allow users to opt-out (B stream)
  - [x] **Send to Tracker email address**

lcampbell2:
Do not send emails if only logs are scans and exports

closes #4472

### DIFF
--- a/services/org-footprint/src/database/get-new-audit-logs.js
+++ b/services/org-footprint/src/database/get-new-audit-logs.js
@@ -22,7 +22,9 @@ const getNewAuditLogs = async ({ query, orgKey, days = 1 }) => {
     throw new Error(`Cursor error occurred while trying to find audit logs: ${err}`)
   }
 
-  return auditLogs
+  // do not send out emails if only scans and exports
+  if (auditLogs.length === auditLogs.filter((log) => ['scan', 'export'].includes(log.action))) return []
+  else return auditLogs
 }
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


services/org-footprint/src/database/get-new-audit-logs.js
```diff
@@ -22,7 +22,9 @@ const getNewAuditLogs = async ({ query, orgKey, days = 1 }) => {
     throw new Error(`Cursor error occurred while trying to find audit logs: ${err}`)
   }
 
-  return auditLogs
+  // do not send out emails if only scans and exports
+  if (auditLogs.length === auditLogs.filter((log) => ['scan', 'export'].includes(log.action)).length) return []
```
FestiveKyle, line 26: Totally fine as is but we could do something like 

```
if ( auditLogs.some((log) => ['add', 'delete', <other inclusions>].includes(log.action)))
```

FestiveKyle (approved):

PR #4765, opened 2023-08-25
Add wildcard sibling data

### LINKED ISSUE #4180: Handle wildcard DNS records properly
A wildcard DNS record is a record in a DNS zone that will match requests for non-existent domain names. A wildcard DNS record is specified by using a * as the leftmost label (part) of a domain name, e.g. *. example.com.

Remember that there will be actual domains behind a wildcard DNS entries. 

- [ ] When a wildcard is detected. Let the user know.  bellabella.bc.rcmp-grc.gc.ca as *.bc.rcmp-grc.gc.ca

FestiveKyle:
As described.

Closes #4180.

image


frontend/src/guidance/GuidancePage.js
```diff
@@ -158,6 +158,11 @@ function GuidancePage() {
             <Trans>Scan Pending</Trans>
           </Badge>
         )}
+        {data.findDomainByDomain.wildcardSibling && (
```
lcampbell2, line 161: Should we put this behind the B stream?

lcampbell2 (changes requested):

Issue #3493, opened 2022-02-22
Collect historical summary data


FestiveKyle:
Currently the scan summary data is only available for the latest scan. We should change how we store summaries to allow for a historical view. 

Current structure (scanSummaries):

```
[
  {
    "_key": "https",
    "_id": "scanSummaries/https",
    "_rev": "_dwC-pyu--B",
    "fail": 1931,
    "pass": 21516,
    "total": 23447
  },
  {
    "_key": "ssl",
    "_id": "scanSummaries/ssl",
    "_rev": "_dwC-qGy--A",
    "fail": 3886,
    "pass": 19561,
    "total": 23447
  },
  {
    "_key": "dkim",
    "_id": "scanSummaries/dkim",
    "_rev": "_dwC-qd---_",
    "fail": 23425,
    "pass": 22,
    "total": 23447
  },
  {
    "_key": "spf",
    "_id": "scanSummaries/spf",
    "_rev": "_dwC-qwS--B",
    "fail": 22890,
    "pass": 557,
    "total": 23447
  },
  {
    "_key": "dmarc",
    "_id": "scanSummaries/dmarc",
    "_rev": "_dwC-rEO--A",
    "fail": 4929,
    "pass": 18518,
    "total": 23447
  }
]
```

Suggested structure:

```
[
  {
    "_key": "123",
    "_id": "scanSummaries/123",
    "_rev": "_dwC-pyu--B",
    "date": "2022-02-22",
    "dmarc": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "dkim": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "spf": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "ssl": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "https": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
  },
  {
    "_key": "124",
    "_id": "scanSummaries/124",
    "_rev": "_dwC-pyu--B",
    "date": "2022-02-23",
    "dmarc": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "dkim": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "spf": {
        "fail": 1931,
        "pass": 21516,
        "total": 23447
    },
    "ssl": {
        "fail": 1931,
        "pass": 21516,
[...]

PR #4772, opened 2023-08-29
Feature/keep historical summary data

lcampbell2:
- Modify how chart summaries are being kept to store each summary in one document with a date (for format see #3493)
- Add hist_summaries value to org docs to gather data for org members
- stop gathering "hidden" summaries


FestiveKyle (approved):
Great stuff :+1:

FestiveKyle (approved):

PR #4815, opened 2023-09-25
Feature/Alert when diff detected in MX records

lcampbell2:
- add notify client to dns-processor
- send personalized email to group of asset subscribers when diff detected


FestiveKyle (changes requested):

scanners/dns-processor/notify/send_mx_diff_email_alerts.py
```diff
@@ -0,0 +1,89 @@
...
+        return {
+            "en": "a host's address was changed",
+            "fr": "l'adresse d'un hôte a été modifiée",
+        }
+    else:
+        return {"en": "unknown", "fr": "inconnue"}
+
+
+def send_mx_diff_email_alerts(
+    domain: str, diff_reason: str, logger: object, db: object
```
FestiveKyle, line 61: `logger` should be `Logger` type, `db` should be `StandardDatabase`

FestiveKyle (approved):
:+1:

PR #4816, opened 2023-09-26
Bugfix/new bilingual invite email

lcampbell2:
- add new email template secret to k8s manifests
- only allow supers to invite new owners


FestiveKyle (approved):
👍

PR #4818, opened 2023-09-26
Fix option to change role to user for super admins

FestiveKyle:
Super admins should have the option to change any user's role to the `USER` role. Currently this option doesn't appear if the user to be changed has the `OWNER` role.

### DIFF
--- a/frontend/src/admin/UserListModal.js
+++ b/frontend/src/admin/UserListModal.js
@@ -254,7 +254,8 @@ export function UserListModal({
                     >
                       {editingUserRole === 'PENDING' && <option value="PENDING">{t`PENDING`}</option>}
                       {(['PENDING', 'USER'].includes(editingUserRole) ||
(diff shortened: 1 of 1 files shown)


frontend/src/admin/UserListModal.js
```diff
@@ -254,7 +254,8 @@ export function UserListModal({
                     >
                       {editingUserRole === 'PENDING' && <option value="PENDING">{t`PENDING`}</option>}
                       {(['PENDING', 'USER'].includes(editingUserRole) ||
-                        (['ADMIN', 'OWNER', 'SUPER_ADMIN'].includes(permission) && editingUserRole === 'ADMIN')) && (
+                        (['ADMIN', 'OWNER', 'SUPER_ADMIN'].includes(permission) && editingUserRole === 'ADMIN') ||
```
lcampbell2, line 257: the SUPER_ADMIN can be removed from the includes array

lcampbell2 (approved):

PR #4819, opened 2023-09-26
Upgrade mass invite script

FestiveKyle:
Script now checks for any missing information and checks current affiliation status for each requested invite, updating roles instead if affiliation already exists.


lcampbell2 (approved):
LGTM

PR #4860, opened 2023-10-12
Fix mx diff check on newly added domains

FestiveKyle:


scanners/dns-processor/service.py
```diff
@@ -150,7 +153,7 @@ async def reconnected_cb():
         logger.info(f"Connected to NATS at {nc.connected_url.netloc}...")
 
     nc = await nats.connect(
-        error_cb=error_cb,
+        # error_cb=error_cb,
```
lcampbell2, line 156: can we uncomment or delete?

lcampbell2 (changes requested):

lcampbell2 (approved):

PR #4912, opened 2023-10-24
Fix root cert color when self-signed

### LINKED ISSUE #4259: Check if invidiual certificates are included in trust store (check for root certs)

FestiveKyle:
Adds `passedValidation` property to certificate chain results. Only set shown certificates as red on frontend if they are self-signed and do not pass validation.

Closes #4259


lcampbell2 (approved):
LGTM :+1:

PR #4934, opened 2023-11-02
Move login buttons not shrink, wrap instead

### LINKED ISSUE #4931: Fix log in button shrinking
Log in buttons can shrink in size making their text partially disappear.

image

FestiveKyle:
Wrap login buttons instead of making them shrink.

Closes #4931

Before:
image

After:
image
image


lcampbell2 (approved):
Nice!

PR #4938, opened 2023-11-03
Bugfix/prevent nulls from killing progress reports

lcampbell2:
- generalize return value names
- exclude null values from avg calcs


FestiveKyle (approved):
We should find where the nulls are coming from - there might be a null for one property but a proper diff in another. We'd be skipping this diff.

PR #5032, opened 2024-01-04
Add progress report graph for orgs

lcampbell2:
- add multi-line chart for last 30 days of org score progress
- can see progress of percentages and domain count
- can switch between summary tiers

image


FestiveKyle (approved):
:clap:

PR #5148, opened 2024-03-01
Include errors for mx and ns checks

FestiveKyle:
Guidance page currently crashes if mx or ns checks gave error as `warnings` will be null. This PR fixes that issue and shows the errors if they occur.

image

### DIFF
--- a/frontend/src/client.js
+++ b/frontend/src/client.js
@@ -105,8 +105,7 @@ const refreshTokens = async () => {
 }
 
 const httpLink = createHttpLink({
-  uri:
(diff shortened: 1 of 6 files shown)


frontend/src/client.js
```diff
@@ -105,7 +105,8 @@ const refreshTokens = async () => {
 }
 
 const httpLink = createHttpLink({
-  uri: process.env.NODE_ENV === 'production' ? `https://${window.location.host}/graphql` : '/graphql',
+  uri:
+    process.env.NODE_ENV === 'production' ? `https://${window.location.host}/graphql` : 'http://localhost:4000/graphql',
```
lcampbell2, line 109: Is this on purpose?

lcampbell2 (approved):

PR #5288, opened 2024-05-07
set TLS statuses to fail when endpoint is present and certificate is missing

### LINKED ISSUE #5283: Count lack of certificate as negative finding
If a web server is present and there is no TLS certificate detected, the INFO status is given to all TLS related statuses. They should be marked as fail for being missing.

lcampbell2:
- gives feedback to users on missing certificate
- evens out summary totals

closes #5283

### DIFF
--- a/scanners/web-processor/web_processor/web_processor.py
+++ b/scanners/web-processor/web_processor/web_processor.py
@@ -10,7 +10,7 @@ guidance_file = open(f"{current_directory}/tls-guidance.json")
 guidance = json.load(guidance_file)
 
 
-def process_tls_results(tls_results):
+def process_tls_results(tls_results, web_server_present):
     neutral_tags = []
     positive_tags = []
     negative_tags = []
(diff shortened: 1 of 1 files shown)


FestiveKyle (changes requested):
`tls_results` will have an error of `ERROR_NO_CONNECTIVITY` if connection couldn't be created. The error check is early in `process_tls_results` and early returns, so we'll have to catch this case at/before then

scanners/web-processor/web_processor/web_processor.py
```diff
@@ -24,6 +24,18 @@ def process_tls_results(tls_results):
     curve_status = "info"
 
     if tls_results.get("error"):
+        # if endpoint is live and no certificate detected, set status to fail
+        if (
+            tls_results.get("scan_status", None) == "NO_NETWORK_CONNECTIVITY"
```
FestiveKyle, line 29: I think this `scan_status` should be `ERROR_NO_CONNECTIVITY`

FestiveKyle (changes requested):

FestiveKyle (approved):

PR #5290, opened 2024-05-08
Increase connect timeout to 10

FestiveKyle:
Some scans fail to connect during the initial connection check but then connect on later attempts. Maybe cold starts? Lets try increasing the timeout.


tahahoussari (approved):
Looks good to me

PR #5292, opened 2024-05-08
Upgrade checkdmarc and ignore unrelated records at _dmarc

### LINKED ISSUE #5269: DMARC results not returned if unrelated records found
DMARC results scan returns an error without data if unrelated records found under _dmarc subdomain. 

We should manually call `check_dmarc` with `ignore_unrelated_records=True`. This call is currently handled in `check_domains`, but it should be easy to remove `check_domains` and just call the individual checks.

FestiveKyle:
Upgrades checkdmarc to new major version and ignores unrelated records at the _dmarc subdomain when checking for DMARC records. Currently a bug exists in checkdmarc if unrelated records exist at _dmarc but there is no record found - gives a `list index out of range` error. I created an issue and PR over there to fix it.

Closes #5269

### DIFF
--- a/scanners/dns-scanner/dns_scanner/email_scanners.py
+++ b/scanners/dns-scanner/dns_scanner/email_scanners.py
@@ -71,7 +71,6 @@ class DMARCScanner:
 
         try:
             # Perform "checkdmarc" scan on provided domain.
-            # scan_result = json.loads(json.dumps(check_domains(domain_list, skip_tls=True, timeout=5.0)))
             scan_result = {
                 "domain": self.domain,
                 "base_domain": get_base_domain(self.domain),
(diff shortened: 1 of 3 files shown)


scanners/dns-scanner/dns_scanner/email_scanners.py
```diff
@@ -70,7 +71,16 @@ def run(self):
 
         try:
             # Perform "checkdmarc" scan on provided domain.
-            scan_result = json.loads(json.dumps(check_domains(domain_list, skip_tls=True, timeout=5.0)))
+            # scan_result = json.loads(json.dumps(check_domains(domain_list, skip_tls=True, timeout=5.0)))
```
lcampbell2, line 74: remove unused code

lcampbell2 (changes requested):

lcampbell2 (approved):

PR #5299, opened 2024-05-16
Reduce connect timeout for TLS Scanner

FestiveKyle:
Reduces the timeout for TLS scanner to 2sec from 10sec. Also raises the connect timeout for the endpoint scanner to 2sec to align with the TLS scanner.


tahahoussari (approved):
LGT

PR #5321, opened 2024-05-23
Bugfix/change nats channel for easm disco service

lcampbell2:
- prevents service from being run during mass scans
- cleans up error messages

### DIFF
--- a/azure-defender-easm/add-domain-to-easm/service.py
+++ b/azure-defender-easm/add-domain-to-easm/service.py
@@ -7,7 +7,6 @@ import nats
 import functools
 import json
 import signal
-import traceback
 
 from clients.easm_client import run_disco_group, create_disco_group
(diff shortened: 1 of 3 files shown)


azure-defender-easm/add-domain-to-easm/service.py
```diff
@@ -20,7 +19,7 @@
 logger = logging.getLogger()
 
 NAME = os.getenv("NAME", "add-domain-to-easm")
-SUBSCRIBE_TO = os.getenv("SUBSCRIBE_TO", "domains.*")
```
FestiveKyle, line 23: This is set in the `deployment.yaml` too so we have to make the change there as well

lcampbell2: Good catch

FestiveKyle (approved):

PR #5340, opened 2024-05-29
Rework dmarc report cosmos loaders

FestiveKyle:
This PR batches the summary queries for cosmosdb together to reduce the number of connections created. Current each domain has leads each date of the 13 dates separately with each date having 5 calls to different parts of the document. This PR will batch 60 domains together into one call. 

We should expect a large speed increase as well - running the service in a dev environment:
- Current state: 9min30s
- After these changes: 0m38s


lcampbell2 (approved):
LGTM, great work on the time optimization!

PR #5351, opened 2024-05-31
Feature/import easm additional findings

### LINKED ISSUE #3182: Server isn't using compression
Istio doesn't offer a native way to configure gzip compression, leaving users directly configuring the [Envoy proxy](https://www.envoyproxy.io/) server. We were never able to figure out a way of enabling gzip that didn't also break Tracker's websockets based one-time-scan functionality so we ended up disabling gzip entirely.
The result is a pretty mediocre performance score, with an obvious solution:
Screenshot from 2021-11-09 10-14-19
Options to consider: 
* reenable gzip, rework the API subscriptions to use [SSE](https://github.com/enisdenjo/graphql-sse) instead of websockets
* Make yet another attempt to create an EnvoyFilter that performs gzip compression without breaking websockets. 

The first option [looks like the best path forward](https://the-guild.dev/blog/graphql-over-sse).

lcampbell2:


azure-defender-easm/import-easm-additional-findings/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+#===============================================================================================
+#===============================================================================================
+
+FROM python:3.11.6-slim-bookworm
+
+# Copy local code to the container image.
+ENV PYTHONUNBUFFERED 1
+ENV PYTHONWARNINGS ignore
+
+WORKDIR /label-known-easm-assets
```
FestiveKyle, line 31: Not a big deal but this name is for a different service

azure-defender-easm/import-easm-additional-findings/cloudbuild.yaml
```diff
@@ -0,0 +1,34 @@
+steps:
+  - name: "gcr.io/cloud-builders/docker"
+    id: generate-image-name
+    entrypoint: "bash"
+    dir: azure-defender-easm/import-easm-additional-findings
+    args:
+      - "-c"
+      - |
+        echo "gcr.io/$PROJECT_ID/azure-defender-easm/import-easm-additional-findings:$BRANCH_NAME-$SHORT_SHA-$(date +%s)" > /workspace/imagename
```
FestiveKyle, line 9: Does this need the `sed` substitution added?

FestiveKyle (changes requested):

FestiveKyle (approved):

PR #5357, opened 2024-06-03
Remove ability to change selectors

FestiveKyle:


FestiveKyle:
Before merging, make following changes:
- Update /Guidance section for DKIM selector changes (no adding/editing/removing)
- Add temporary banner with similar info from /Guidance DKIM selector section

frontend/src/app/ReadGuidancePage.js
```diff
@@ -437,14 +438,20 @@ export default function ReadGuidancePage() {
               </ListItem>
             </UnorderedList>
           </ListItem>
-          <ListItem>
+          {/* 6 */}
+          <ListItem id="dkimSelectors">
```
lcampbell2, line 442: since this is used in the url, can we change this id to kebab case?

lcampbell2 (changes requested):

lcampbell2 (approved):

PR #5380, opened 2024-06-10
Bugfix: fix logging in easm label service

lcampbell2:


azure-defender-easm/label-known-easm-assets/service.py
```diff
@@ -11,6 +11,11 @@
 
 load_dotenv()
 
+logging.basicConfig(
+    level=logging.info, format="[%(asctime)s :: %(name)s :: %(levelname)s] %(message)s"
+)
```
FestiveKyle, line 16: Level should use the `logging.INFO` level. Should call `basicConfig` before the client imports so they can use it too.

FestiveKyle (approved):

FestiveKyle (changes requested):
Oops, meant to request changes. See last comment

FestiveKyle (approved):

PR #5381, opened 2024-06-10
Canned message and request invite button

tahahoussari:
Added an error message when the user tries to view the scan results of a domain when they don't have permission. Lists organisations that the domain belongs to so that the user can request access to the needed organisation. Added a couple test cases.


FestiveKyle (approved):
LGTM. Well done 🔥

PR #5436, opened 2024-06-27
Security: Add network policies

### LINKED ISSUE #3774: NetworkPolicies
Back in eb395ca we enabled a network policy. It worked on GKE, but blocked database access on AKS. We reverted it in e9c72bb and never had time to go back and fix it, in part because while [debugging NetworkPolicy issues on GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy-logging#accessing_logs) is quite simple the same cannot be said for AKS.

Revisiting this is needed but will need to take into consideration the [similarities/differences between Istio's policies and network policies](https://istio.io/latest/blog/2017/0.1-using-network-policy/) as well as the fact that the [CNI plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/) used on [AKS is Calico](https://docs.microsoft.com/en-us/azure/aks/use-network-policies#network-policy-options-in-aks) while it's[Cilium on GKE](https://cilium.io/blog/2020/08/19/google-chooses-cilium-for-gke-networking).

lcampbell2:
Add Kubernetes network policies to restrict ingress and egress between pods to only expected traffic:
- Create netpols for the frontend, api, and arangodb pods.
- Labels have been added to pods so they can be allowed to send/receive traffic.
- Create netpols to default-deny pod access to VM instance metadata on AKS.

closes #3774

### DIFF
--- a/k8s/infrastructure/bases/arangodb/network-policy.yaml
+++ b/k8s/infrastructure/bases/arangodb/network-policy.yaml
@@ -41,6 +41,67 @@ spec:
           podSelector:
             matchLabels:
               app: web-processor
+        # additonal services
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: scanners
+          podSelector:
+            matchLabels:
+              app: domain-discovery
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: scanners
+          podSelector:
+            matchLabels:
+              app: summaries
+        - namespaceSelector:
+            matchLabels:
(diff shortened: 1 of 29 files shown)


k8s/infrastructure/bases/arangodb/network-policy.yaml
FestiveKyle, line 1: We need to include the dns-scanner and domain-dispatcher for the scans to run

FestiveKyle (changes requested):

FestiveKyle (approved):

PR #5440, opened 2024-07-03
filter imported WebComponentCves to top 25 cves

lcampbell2:
filter cves imported from easm to only top25 entries


tahahoussari (approved):
LGTM

PR #5466, opened 2024-07-12
Bugfix: Conditionally use DKIM status for calculating mail summary

lcampbell2:
Will only use non-info DKIM statuses when calculating mail summaries


tahahoussari (approved):
LGTM

tahahoussari (approved):
LGTM

PR #5475, opened 2024-07-17
Security: Add security context to pod deployments

lcampbell2:
adds `securityContext` to k8s pod deployments configuring the following options:
- Discretionary Access Control: Permission to access an object, like a file, is based on [user ID (UID) and group ID (GID)](https://wiki.archlinux.org/index.php/users_and_groups).
- Running as privileged or unprivileged.
- [Linux Capabilities](https://linux-audit.com/linux-capabilities-hardening-linux-binaries-by-removing-setuid/): Give a process some privileges, but not all the privileges of the root user.
- [`Seccomp`](https://kubernetes.io/docs/tutorials/security/seccomp/): Filter a process's system calls.
- `allowPrivilegeEscalation`: Controls whether a process can gain more privileges than its parent process. 
- `readOnlyRootFilesystem`: Mounts the container's root filesystem as read-only.


k8s/apps/bases/frontend/deployment.yaml
```diff
@@ -29,4 +33,14 @@ spec:
             requests:
               cpu: 200m
               memory: 45Mi
+          securityContext:
+            runAsGroup: 1337
+            runAsUser: 1337
```
FestiveKyle, line 38: istio apparently doesn't like sharing UIDs. Works for us right now, but something something iptables something something outgoing traffic

k8s/apps/bases/scanners/dns-scanner/deployment.yaml
```diff
@@ -65,4 +69,14 @@ spec:
...
           resources: {}
+          securityContext:
+            runAsGroup: 1000
+            runAsUser: 1000
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            privileged: false
+            readOnlyRootFilesystem: true
```
FestiveKyle, line 80: We save the PSL cache in /tmp so we need that dir writable

FestiveKyle (changes requested):

FestiveKyle (approved):

PR #5476, opened 2024-07-17
Feature: Add IPs to domains export CSV

### LINKED ISSUE #5424: Add IP to CSV export
It would be useful to include the IP(s) in the csv file shown in the Web Guidance section.

For example, our CMDB is not complete, this would help building it.

lcampbell2:
grabs domain IP addresses from webScans if available and adds them as semicolon-separated list

closes #5424

### DIFF
--- a/api/src/organization/objects/organization.js
+++ b/api/src/organization/objects/organization.js
@@ -159,6 +159,7 @@ export const organizationType = new GraphQLObjectType({
         })
         const headers = [
           'domain',
+          'ipAddresses',
           'https',
           'hsts',
           'certificates',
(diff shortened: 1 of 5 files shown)


api/src/organization/objects/organization.js
```diff
@@ -177,7 +178,8 @@ export const organizationType = new GraphQLObjectType({
         let csvOutput = headers.join(',')
         domains.forEach((domain) => {
           let csvLine = `${domain.domain}`
-          csvLine += headers.slice(1, 10).reduce((previousValue, currentHeader) => {
+          csvLine += `,${domain.ipAddresses.join(';')}`
```
FestiveKyle, line 181: we're using pipe as a separator for tags, can we use it here too?

api/src/organization/queries/get-all-organization-domain-statuses.js
```diff
@@ -48,7 +49,14 @@ export const getAllOrganizationDomainStatuses = {
     ]
     let csvOutput = headers.join(',')
     domainStatuses.forEach((domainStatus) => {
-      const csvLine = headers.map((header) => `"${domainStatus[header]}"`).join(',')
+      const csvLine = headers
+        .map((header) => {
+          if (header === 'ipAddresses') {
+            return domainStatus[header].join(';')
```
FestiveKyle, line 55: we're using pipe as a separator for tags, can we use it here too?

FestiveKyle (changes requested):

PR #5543, opened 2024-08-16
Fix user search

### LINKED ISSUE #5206: Expose '%' to users for searching domains, users, etc
Improve search for 

- [ ] users
- [x] domains
- [x] orgs

Allow other operators

- [x] %
- [ ] * (same as %)

FestiveKyle:
Fixes the following problems for user searching:
  - Currently same users are shown multiple times if multiple search terms matched (especially noticeable during email searches)
  - Email searches are tokenized in a way that breaks searching
    - Searching `user@some-domain.gc.ca` creates a search for the individual terms: `user`, `some`, `domain.gc.ca`

Switch email searching to be a substring search.

Closes #5206


tahahoussari (approved):
LGTM

PR #5621, opened 2024-09-11
Update Top 25 Labels

lcampbell2:
- restyle DomainCard to handle many badges and tags
- update CVE_DETECTED label to 'SPIN Top 25' 
image


FestiveKyle (changes requested):
I'm getting overflow problems on long domains:

image

FestiveKyle (approved):

PR #5713, opened 2024-09-20
Add ability to ignore/unignore cves per domain

FestiveKyle:
Add api/frontend ability to ignore/unignore cves per domain. Does not affect resource queries, just sets state for now.

Next:
- Actually use these ignoredCves where required
- Add appropriate audit logs (only for super admins vs all verified orgs with claims to domain)


lcampbell2 (changes requested):
Can the ignore cve mutation code be moved into it's own component from AdditionalFindings.js? That file is already pretty bloated.

FestiveKyle:
> Can the ignore cve mutation code be moved into it's own component from AdditionalFindings.js? That file is already pretty bloated.

It is done

lcampbell2 (approved):

Issue #5758, opened 2024-10-02
Remove user preferred language


FestiveKyle:
- [x] Remove field when creating user
- [x] Remove field on account page
- [x] Remove usage when sending 2FA message
  - [x] Updating phone number
  - [x] Updating email

PR #5771, opened 2024-10-04
Remove preferred lang

FestiveKyle:
Remove features to add/change preferred language and any usage of preferred language.

Reworks username update flow from:
- Change username in UI
- Username gets updated in DB. Email validation is changed to false.
- Email validation email is sent. 
- Following link sets email validation to true.

To:
- Change username in UI
- Username not updated. Email validation is not changed.
- Email validation email is sent.
- Following link sets username to new email and sets validation to true.

Adds an "Email Updated" email which is sent to user's previous email address upon completing/validating email change.

Closes [#5758](https://github.com/canada-ca/tracker/issues/5758)


frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "tracker-frontend",
   "version": "0.1.0",
   "scripts": {
-    "build": "webpack --env production --config webpack.config.js",
+    "build": "webpack --env development --config webpack.config.js",
```
lcampbell2, line 5: should stay production

lcampbell2 (changes requested):

frontend/package.json
FestiveKyle: Oof good one

lcampbell2 (approved):

PR #5800, opened 2024-10-10
Feature: Add total negative findings to org summaries

lcampbell2:


services/summaries/summaries.py
```diff
@@ -58,6 +58,33 @@ def ignore_domain(domain):
...
+            )
+            FOR tag IN FLATTEN([emailTags, webTags], 2)
+                FILTER tag != null
+                RETURN tag
+        """,
+        bind_vars={"domain_id": domain_id},
+    )
+    if cursor.empty():
+        return []
+    return cursor.next()
```
FestiveKyle, line 85: We're only returning one tag here

services/summaries/summaries.py
```diff
@@ -58,6 +58,33 @@ def ignore_domain(domain):
...
+                    SORT dnsScan.timestamp DESC
+                    LIMIT 1
+                    RETURN FLATTEN([dnsScan.dmarc.negativeTags, dnsScan.dkim.negativeTags, dnsScan.spf.negativeTags])
+            )[0]
+            LET webTags = (
+                FOR web, webE IN 1 OUTBOUND @domain_id domainsWeb
+                    SORT web.timestamp DESC
+                    LIMIT 1
+                    FOR webScan, webScanE IN 1 OUTBOUND web webToWebScans
+                        RETURN FLATTEN([webScan.results.tlsResult.negativeTags, webScan.results.connectionResults.negativeTags])
```
FestiveKyle, line 75: This will count web tags multiple times per domain if the endpoints each have the tag. Which is fine if we want to display it that way. 
```
https14: 321 domains
```
 vs 
```
https14: 400 web servers/endpoints/whateverweshouldcallit
```

FestiveKyle (changes requested):

services/summaries/summaries.py
lcampbell2: I think we do, since it's a summary of the total amount of negative tags. Good catch - I'll work on the wording

FestiveKyle (approved):

PR #5804, opened 2024-10-11
Handle ignored cves

FestiveKyle:
- [x] Exclude ignored CVEs from AdditionalFindings (frontend only)
- [x] Exclude ignored CVEs from CSV export
- [x] Log action to every verified org with claim to domain
  - [x] Log as an "UPDATE" action with updated properties of "name: <CVE-NAME>, "oldValue": "ignored", "newValue": "unignored"


frontend/src/guidance/IgnoredCves.js
```diff
@@ -1,47 +1,96 @@
...
+            <Trans>Undetected Ignored CVEs:</Trans>
+          </Text>
+          <Box px="2">
+            {!undetectedIgnoredCves || undetectedIgnoredCves.length === 0 ? (
+              <Text>
+                <Trans>None</Trans>
+              </Text>
+            ) : (
+              <SimpleGrid columns={8}>
+                {undetectedIgnoredCves &&
```
lcampbell2, line 70: empty or null array already checked for

frontend/src/guidance/IgnoredCves.js
```diff
@@ -51,6 +100,7 @@ function IgnoredCves({ ignoredCves, setActiveCveHandler }) {
 export default withSuperAdmin(IgnoredCves)
 
 IgnoredCves.propTypes = {
-  ignoredCves: PropTypes.array.isRequired,
+  undetectedIgnoredCves: PropTypes.arrayOf(PropTypes.string.isRequired).isRequired,
+  detectedIgnoredCves: PropTypes.object.isRequired,
   setActiveCveHandler: PropTypes.func.isRequired,
```
lcampbell2, line 105: is it better to import the whole object or the individual prop types? if no performance/memory advantage this is fine

lcampbell2 (changes requested):

frontend/src/guidance/IgnoredCves.js
FestiveKyle: I don't think there would be any performance loss either way. We already have these separated and the `undetectedIgnoredCves` are just strings since we aren't getting the vuln level from the api.

lcampbell2 (approved):

PR #6078, opened 2025-01-23
Feature: Add external ID field to orgs

lcampbell2:
- optional field for orgs
- users can read
- supers can update
- exported in CSV


frontend/src/admin/OrganizationInformation.js
```diff
@@ -318,6 +319,7 @@ export function OrganizationInformation({ orgSlug, removeOrgCallback: setSelecte
                   <Box gridColumn={{ base: 'span 4', md: 'span 2' }} mb="0.5rem">
                     <FormField name="cityFR" label={t`City (FR)`} />
                   </Box>
+                  {externalIdField}
```
FestiveKyle, line 322: Getting an error here for function not being a valid React child. Should probably also move the `mb` from City (FR) Box as well to keep the layout.

FestiveKyle (changes requested):

FestiveKyle (approved):

PR #6692, opened 2025-10-01
Change logic for decay check to work with dynamic min scans var 

saraj17:


lcampbell2 (approved):
Nice use of reusable logic and meaningful variable names! LGTM :+1:

PR #7295, opened 2026-04-29
fix: API devcontainer for Codespaces

lcampbell2:
The API devcontainer was broken in Codespaces due to several issues. This PR fixes them:

- docker-in-docker: adds moby: false to the feature config — Moby packages were removed from Debian trixie (base image used by javascript-node:20), causing container build to fail
- Supporting services: removes arangodb/nats from the devcontainer compose entirely; they now start via postStartCommand using the existing api/docker-compose.yaml through docker-in-docker, avoiding healthcheck/startup race conditions that blocked container creation
- Environment: adds all required env vars with dev-safe dummy values so the API starts without manual .env setup
- api/docker-compose.yaml: adds fallback volume (arangodb-data) when TRACKER_BACKUP_DIR is unset, so arangodb works in Codespaces without local config


saraj17 (approved):
LGTM

PR #7409, opened 2026-06-18
security(scanners|services): remove unused PyJWT package from requirements.txt

lcampbell2:


FestiveKyle (changes requested):
PyJWT is a transitive dep of python-arango and notifications-python-client and will still be installed

PR #7442, opened 2026-07-07
fix(dns-scanner): extended CNAME/NS results

lcampbell2:
- Updated `dns_scanner.py` to:
  - include `CNAME` in the `record_exists` gate (alongside `A`, `SOA`, `NS`),
  - preserve/derive `rcode` from multiple query attempts more reliably,
  - add new result fields:
    - `ns_delegations`
    - `registrar_context`
  - call new helper functions for NS delegation checks and RDAP registrar context.

- Added new module `ns_registrar.py` with:
  - `check_ns_delegations(...)`
    - resolves NS hosts,
    - probes authoritative SOA responses directly against nameserver IPs,
    - classifies delegation health as `none` / `partial` / `full` lame delegation.
  - `get_registrar_context(...)`
    - performs RDAP lookup via `https://rdap.org/domain/<domain>`,
    - extracts registrar metadata and RDAP nameservers,
    - compares observed delegation NS hosts with RDAP NS data.


scanners/dns-scanner/dns_scanner/ns_registrar.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def probe_nameserver(
+    where: str, qname: str, qtype: str, recursion_desired: bool, timeout: int
+):
+    query = dns.message.make_query(
+        qname,
+        dns.rdatatype.from_text(qtype),
+        use_edns=True,
+        recursion_desired=recursion_desired,
```
FestiveKyle: `make_query()` doesn't seem to have a `recursion_desired` parameter, raising TypeError on call instead (and defaulting to `lame_type=full` for all scans)

scanners/dns-scanner/dns_scanner/ns_registrar.py
```diff
@@ -0,0 +1,195 @@
...
+def check_ns_delegations(domain, zone_apex, ns_records, resolver=None, timeout_sec=10):
+    if resolver is None:
+        resolver = dns.resolver.get_default_resolver()
+
+    qname = zone_apex
+    if not zone_apex:
+        qname = domain
+
+    ns_hosts = ns_records.get("hostnames", [])
+    if len(ns_records) == 0:
```
FestiveKyle: I think this will always be false, `ns_records`  comes from checkdmarc and should always have the shape `{'hostnames': [...], 'warnings|errors': [...]}`

scanners/dns-scanner/dns_scanner/dns_scanner.py
```diff
@@ -201,14 +213,19 @@ def scan_domain(domain, dkim_selectors=None):
...
     # Check if domain exists
     dns_answer_return_types = []
-    for query_type in [dns.rdatatype.A, dns.rdatatype.SOA, dns.rdatatype.NS]:
+    query_res = {}
+    for query_type in [
+        dns.rdatatype.A,
+        dns.rdatatype.SOA,
+        dns.rdatatype.NS,
+        dns.rdatatype.CNAME,
+    ]:
```
FestiveKyle: This section is intended to be a gate before the full scan. If a `NOERROR` is returned it breaks early and continues onto the scans (skipping the other gate queries as some domains have pretty slow responses). Ideally would keep this section as-is, keeping the break, and not use it for collecting real data.

scanners/dns-scanner/dns_scanner/ns_registrar.py
```diff
@@ -0,0 +1,195 @@
...
+            ns_aaaa = resolver.resolve(host, "AAAA")
+            if ns_aaaa:
+                ns_ip = ns_aaaa[0].to_text()
+        except (NoAnswer, NXDOMAIN, NoNameservers, Timeout):
+            ns_ip = None
+
+    return ns_ip
+
+
+def check_ns_delegations(domain, zone_apex, ns_records, resolver=None, timeout_sec=10):
```
FestiveKyle, line 59: I'm having trouble getting this to flag a real lame nameserver, do you have a working domain I can test it against?

scanners/dns-scanner/dns_scanner/ns_registrar.py
lcampbell2: my bad, asked opencode to move the logic to a separate file and didn't check for hallucinations

scanners/dns-scanner/dns_scanner/ns_registrar.py
lcampbell2: same as above

scanners/dns-scanner/dns_scanner/dns_scanner.py
lcampbell2: I'll find a compromise

PR #7470, opened 2026-07-24
Add experimental PQC scanning support

FestiveKyle:


scanners/web-processor/web_processor/web_processor.py
```diff
@@ -1,7 +1,12 @@
 import os
 import json
+import logging
 from urllib.parse import urlparse
 
+logger = logging.getLogger(__name__)
+
+PQC_SCAN_ENABLED = os.getenv("PQC_SCAN_ENABLED", "false").lower() in ("true", "1")
```
lcampbell2, line 8: this value is missing from the web-processor deployment manifest, so will always be false

api/src/web-scan/objects/experimental-result.js
```diff
@@ -0,0 +1,86 @@
...
+      type: GraphQLBoolean,
+      description: `Whether the scanned server accepted at least one post-quantum hybrid key exchange group.`,
+      resolve: async ({ supportsPqKeyExchange }) => supportsPqKeyExchange,
+    },
+    supportedPqGroups: {
+      type: new GraphQLList(GraphQLString),
+      description: `The post-quantum hybrid key exchange groups accepted by the scanned server. Null when the server does not support TLS 1.3, in which case nothing was tested.`,
+      resolve: async ({ supportedPqGroups }) => supportedPqGroups,
+    },
+    tls13Supported: {
```
lcampbell2: web-processor converts this to camel case as "tls1-3Supported"

lcampbell2 (changes requested):

scanners/web-processor/web_processor/web_processor.py
FestiveKyle: Good catch, added

api/src/web-scan/objects/experimental-result.js
FestiveKyle: Done, switched back to the same style as the TLS field names in `tls-result.js`

lcampbell2 (approved):

PR #7518, opened 2026-08-14
feat(scanners): additional findings processor

lcampbell2:
Adds a new scanners/findings-processor Go service that consumes finding events from NATS JetStream and upserts normalized documents into ArangoDB additionalFindings.
- Introduces a standalone service scaffold (cmd/service, internal/*, go.mod, Dockerfile, Makefile, cloudbuild, env template, README).
- Implements config/env loading, structured logging, and graceful shutdown.
- Subscribes to scans.findings.* (configurable), validates event payloads, and applies explicit Ack/Nak/Term handling.
- Adds finding normalization + deterministic key generation, then upsert logic that creates new docs or updates lastSeen and occurrence count for existing docs.
- Includes broad unit test coverage across config, model validation/mapping, DB upsert flow, and runner/event behavior.


scanners/findings-processor/internal/database/upsert.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+
+	"github.com/arangodb/go-driver/v2/arangodb"
+	"github.com/canada-ca/tracker/scanners/findings-processor/internal/model"
+	"github.com/rs/zerolog/log"
+)
+
+type findingUpdatePatch struct {
+	LastSeen       string `json:"lastSeen"`
+	OccurenceCount int    `json:"occurenceCount"`
```
FestiveKyle: Typo occurenceCount/occurrenceCount leads to duplicate document fields and API query failure

scanners/findings-processor/internal/database/upsert.go
```diff
@@ -0,0 +1,86 @@
...
+		var finding model.FindingDocument
+		err := readFindingDocument(ctx, findingsCol, key, &finding)
+		if err != nil {
+			log.Warn().Err(err).Msg("failed to read finding doc")
+			return err
+		}
+
+		patch := findingUpdatePatch{
+			LastSeen:       evt.ObservedAt,
+			OccurenceCount: finding.OccurrenceCount + 1,
```
FestiveKyle: Typo occurenceCount/occurrenceCount leads to duplicate document fields and API query failure

scanners/findings-processor/internal/model/finding.go
```diff
@@ -0,0 +1,116 @@
...
+	Raw             map[string]any `json:"raw"`
+	Status          string         `json:"status"`
+}
+
+func (e FindingEvent) GetKey() string {
+	// domainKey + source + findingType + subject + reasonCode
+	keyArgs := []string{e.DomainKey, e.Source, e.FindingType, e.Subject, e.ReasonCode}
+	data := strings.Join(keyArgs, "_")
+	return b64.StdEncoding.EncodeToString([]byte(data))
+}
```
FestiveKyle, line 60: The derived key seems like it could be problematic. It is broken in the API (#7534) as it expects a typical incremented number:
```
relayAfterTemplate = aql`FILTER TO_NUMBER(finding._key) > TO_NUMBER(${cursorList[0].id})`
```
One idea to fix, use auto incrementing _key and save the unique derived key as a separate field which can be indexed. We can keep the generation the same, (though the joining character should be changed if `_` is expected in any of the args. `SPF_MISSING` is in finding_test.go for example), or we could also hash the result and store that (base64 probably shouldn't be used as it can include `/` which isn't allowed as an arangodb key).

scanners/findings-processor/internal/runner/actions.go
```diff
@@ -0,0 +1,22 @@
...
+	switch action {
+	case "ack":
+		_ = msg.Ack()
+	case "nak":
+		_ = msg.Nak()
+	case "term":
+		_ = msg.Term()
+	default:
+		_ = msg.Term()
+	}
```
FestiveKyle: Could this be removed and the msg actions just be called directly instead? This removes intellisense for me and defaulting to a `.Term()` seems risky.

scanners/findings-processor/internal/database/upsert_test.go
FestiveKyle, line 1: Not sure of the effectiveness of these tests, they mainly just checks error paths - given a failure, the function errors. Some integration tests with a live test DB would be beneficial.

scanners/findings-processor/internal/database/upsert.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		log.Warn().Err(err).Msg("failed to find collection")
+		return err
+	}
+
+	key := evt.GetKey()
+
+	findingExists, err := documentExists(ctx, findingsCol, key)
+	if err != nil {
+		log.Warn().Err(err).Msg("failed to find collection")
```
FestiveKyle: Should this be something like "failed to find document"?

scanners/findings-processor/README.md
```diff
@@ -0,0 +1,73 @@
...
+- Subscribes to `scans.findings.*` (configurable)
+- Validates incoming payloads
+- Writes findings to ArangoDB
+- Acknowledges messages with explicit `Ack` / `Nak` / `Term` behavior
+
+## Event contract (current)
+
+Expected JSON fields:
+
+- Required: `schemaVersion`, `source`, `findingType`, `domainKey`, `subject`, `confidence`, `observedAt`
```
FestiveKyle: `schemaVersion` doesn't seem to be used

scanners/findings-processor/internal/runner/deps.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		getDatabase: func(ctx context.Context, client dbClient, name string) (arangodb.Database, error) {
+			return client.GetDatabase(ctx, name, nil)
+		},
+		drainSubscription: func(sub *nats.Subscription) error {
+			if sub == nil {
+				return nil
+			}
+			return sub.Drain()
+		},
```
FestiveKyle: The description says this does a graceful shutdown but I think sub.Drain() returns immediately. Might have to wait on a `nats.ClosedHandler` for a graceful shutdown.

scanners/findings-processor/internal/runner/processor.go
```diff
@@ -0,0 +1,140 @@
...
+	dbCtx, cancelDB := context.WithTimeout(ctx, 10*time.Second)
+	defer cancelDB()
+
+	db, err := client.GetDatabase(dbCtx, cfg.DBName, nil)
+	if err != nil {
+		return fmt.Errorf("get database failed: %w", err)
+	}
+
+	handler := func(msg *nats.Msg) {
+		eventCtx, cancelEvent := context.WithTimeout(ctx, 5*time.Second)
```
FestiveKyle: I think this should use a `context.Background()` instead of the one derived from the interrupt handler, otherwise Arango might cancel in UpsertFinding.

scanners/findings-processor/internal/runner/processor.go
```diff
@@ -0,0 +1,140 @@
...
+		default:
+			_ = msg.Nak()
+		}
+	}
+
+	var sub *nats.Subscription
+	if cfg.NATSQueueGroup != "" {
+		sub, err = js.QueueSubscribe(cfg.NATSSubject, cfg.NATSQueueGroup, handler, opts...)
+	} else {
+		sub, err = js.Subscribe(cfg.NATSSubject, handler, opts...)
```
FestiveKyle: The underlying durable consumers will be deleted during the drain if they were created while subscribing. Should be able to avoid it by using `CreateOrUpdateConsumer` like the subdomain takeover scanner

scanners/findings-processor/internal/database/upsert.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if _, err := cursor.ReadDocument(ctx, &finding); err != nil {
+		return nil, err
+	}
+	return &finding, nil
+}
+
+func UpsertFinding(ctx context.Context, db arangodb.Database, evt model.FindingEvent) error {
+	findingsCol, err := db.GetCollection(ctx, "additionalFindings", nil)
```
FestiveKyle, line 43: These new findings will have a different shape than the old documents in additionalFindings. What are the plans for the old documents?

scanners/findings-processor/cloudbuild.yaml
```diff
@@ -0,0 +1,43 @@
+steps:
+  - name: "golang:1.25"
+    id: ci-checks
+    dir: scanners/findings-processor
+    entrypoint: "bash"
+    args:
+      - "-c"
+      - |
+        make ci
```
FestiveKyle, line 9: Does this run the integration tests as well?

scanners/findings-processor/internal/database/upsert.go
lcampbell2: old docs will be archived to a separate collection or removed completely - mostly false positive or outdated info

scanners/findings-processor/cloudbuild.yaml
lcampbell2: added separate CI step

scanners/findings-processor/internal/runner/processor.go
lcampbell2: migrated to jetstream consumer

scanners/findings-processor/internal/runner/processor.go
lcampbell2: good catch

scanners/findings-processor/internal/runner/processor.go
```diff
@@ -0,0 +1,121 @@
...
+		if ackErr != nil {
+			log.Warn().Err(ackErr).Msg("failed to ack/nak/term message")
+		}
+	}
+
+	consumeCtx, err := cons.Consume(handler)
+	if err != nil {
+		return fmt.Errorf("failed to create consumer context: %w", err)
+	}
+	defer consumeCtx.Drain()
```
FestiveKyle: 2aa4b80 removed the graceful shutdown logic

scanners/findings-processor/internal/runner/processor.go
```diff
@@ -0,0 +1,121 @@
...
+			ackErr = msg.Term()
+		default:
+			ackErr = msg.Nak()
+		}
+		if ackErr != nil {
+			log.Warn().Err(ackErr).Msg("failed to ack/nak/term message")
+		}
+	}
+
+	consumeCtx, err := cons.Consume(handler)
```
FestiveKyle: Consume errors like a NATS server shutdown or no heartbeat are ignored, logging them in a handler here would be helpful

scanners/findings-processor/go.mod
```diff
@@ -0,0 +1,79 @@
+module github.com/canada-ca/tracker/scanners/findings-processor
+
+go 1.25.0
+
+require (
+	github.com/joho/godotenv v1.5.1
+	github.com/kelseyhightower/envconfig v1.4.0
+	github.com/nats-io/nats.go v1.37.0
```
FestiveKyle: Fairly old version, are newer ones compatible?

scanners/findings-processor/internal/runner/processor.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/canada-ca/tracker/scanners/findings-processor/internal/config"
+	"github.com/canada-ca/tracker/scanners/findings-processor/internal/database"
+	"github.com/canada-ca/tracker/scanners/findings-processor/internal/model"
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
+	"github.com/rs/zerolog/log"
+)
+
+func Run(cfg config.Config) error {
+	nc, err := nats.Connect(cfg.NATSURL)
```
FestiveKyle: We should have some disconnect/reconnect/closed logging handlers here for debugging, similar to the other scanners. Setting `MaxReconnects(-1)` might be a good idea too as right now it just gives up after 2min and sits idle forever

FestiveKyle (commented):
Should be all good after these 👍

FestiveKyle (approved):

PR #7528, opened 2026-08-18
Suspend easm import job

lcampbell2:


FestiveKyle (commented):
Looks like the formatter used for the frontend code didn't pick up the prettier config (should be 2 space + single quotes). Should be good after that

FestiveKyle (approved):

PR #7542, opened 2026-08-26
Frontend/update webpack dev server

FestiveKyle:


frontend/.nvmrc
lcampbell2, line 1: is this file necessary?

FestiveKyle: It lets you call `nvm use` from the directory to activate the correct version of node, I've found it really convenient

https://github.com/nvm-sh/nvm

lcampbell2 (approved):

PR #7580, opened 2026-09-13
fix(frontend): persist UI locale so new tabs keep the selected language

### LINKED ISSUE #5599: New tab opens in default user language even if current language is different
To reproduce:
- Go to any page
- Use language switch to change to non-default language
- Follow any internal link in new tab
- New tab will have user default language, not currently selected language

soroush5:
Fixes #5599.

Switching language with the locale switcher only called `i18n.activate()` in memory — the choice was never stored anywhere. Every new tab boots through `setUpApp`, which activates the server default (or the browser language), so the new tab always came up in the default language instead of the one just selected.

This stores the activated locale in `localStorage` and reads it back at boot, keeping the existing precedence for first-time visitors: stored choice, then `APP_DEFAULT_LANGUAGE`, then the browser language. If storage is unavailable the app behaves exactly as before.

How to test:

1. `npx jest src/utilities/__tests__/i18nPersistence.test.js` in `frontend` — 4 tests, all fail without the fix and pass with it.
2. Manual: switch to Français, open any internal link in a new tab — the new tab stays in French. Verified the same flow in headless Chromium against the real `i18n.config` module: before the fix the second tab read `null`, after the fix it reads `fr`.
3. `src/utilities` + `LocaleSwitcher` suites (27 tests) still pass.


soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #7583, opened 2026-09-15
fix(api|frontend): improve access controls for creating/updating domains with restricted fields

lcampbell2:
- Fix broken RBAC on domain archived/ignoreRua/highAvailability fields: admins could set them on create/update despite access-control.js declaring them super-admin-only, because nothing actually consulted the grant's attrs.
- Add auth.getDeniedFields (api/src/auth/utils/get-denied-fields.js), which uses accesscontrol's own Permission.filter() against the existing grant definitions to detect denied fields, replacing a hand-maintained field list duplicated in both mutations. Wired into context.auth in create-context.js.
- create-domain.js/update-domain.js now call this shared check instead of ad hoc per-field logic (previously only highAvailability was checked on create at all).
- AdminDomainModal.js: only include these fields in createDomain/updateDomain mutation variables when the acting user is super_admin, so non-super-admins no longer trip the new 403 just by submitting an unrelated edit.


frontend/src/admin/AdminDomainModal.js
```diff
@@ -59,6 +59,7 @@ export function AdminDomainModal({
   const toast = useToast()
   const initialFocusRef = useRef()
   const { t } = useLingui()
+  const isSuperAdmin = permission === 'super_admin'
```
FestiveKyle, line 62: I think this will always be false, should check against `SUPER_ADMIN`

FestiveKyle (approved):

PR #7590, opened 2026-09-18
refactor(frontend): additional findings tab

lcampbell2:


frontend/src/guidance/GuidancePage.js
FestiveKyle, line 53: URL path shows `/domains/<domain>/undefined` when navigating to the additional findings tab, new tab should be added to the names

frontend/src/graphql/queries.js
```diff
@@ -647,40 +641,38 @@ export const DOMAIN_GUIDANCE_PAGE = gql`
...
-  query GuidanceAdditionalFindings($domain: DomainScalar!) {
+  query GuidanceAdditionalFindings(
+    $domain: DomainScalar!
+    $orderBy: AdditionalFindingOrder
+    $filters: [AdditionalFindingFilter]
+    $limit: Int
+    $first: Int
+    $after: String
+  ) {
     findDomainByDomain(domain: $domain) {
```
FestiveKyle, line 652: The `id` field should be included here so additionalFindings can be normalized correctly and reduce refetching.

frontend/src/guidance/AdditionalFindings.js
FestiveKyle, line 1: Pagination of additional findings doesn't work currently, need to include the `relayStylePagination()` field in client.js

